PR #14, opened 2015-03-31
Use same scheme when accessing external resources

hexchain:
GitHub pages uses HTTPS, so there's mixed content.


hexchain:
Maybe the same page in `gh-pages` branch needs to be fixed too?

Valve:
Yes, please

PR #20, opened 2015-04-20
fixes webgl fingerprinting

Gheek:
As discussed at gitter, the wrong canvas element (an empty one) is used in the result array of the webgl fingerprint function. Instead the canvas of the (web-)gl object should be used. This commit fixes this issue.


Valve:
Thanks!

Valve:
Sorry, a bit overwhelmed currently, hope to return to the lib soon

PR #32, opened 2015-05-28
Debug flag as option, default false

dario1985:


Valve:
@dario1985, thanks for the PR. I wanted originally to use conditional minification. https://github.com/mishoo/UglifyJS2#conditional-compilation

So that the unminified version will contain all log messages, but the minified version will be compiled without those code branches at all to reduce the size. Your solution will leave the log messages in code even when debug = false

Valve:
Fixed in dd88fb822735863111bcd3976d2e601398e67978

PR #35, opened 2015-06-04
refactor fontsKey, flashFontsKey to do js font detection by default, …

mhuisman:
…falling back to flash font detection if user specifies.  prompted by https://github.com/Valve/fingerprintjs2/issues/23

NOTE: This change is mutative to fingerprints generated with default options.  Pass in options { excludeJsFonts : true } to emulate previous flash fingerprinting behavior.

### DIFF
--- a/fingerprint2.js
+++ b/fingerprint2.js
@@ -176,41 +176,38 @@
       return keys;
     },
     fontsKey: function(keys, done) {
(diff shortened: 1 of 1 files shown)


Valve:
@mhuisman thank you for the PR!

Don't worry about breaking existing FPs, this lib is under development, I made no promise to keep backwards compat yet :)

PR #37, opened 2015-06-06
Add screen.availHeight and screen.availWidth to getScreenResolution

mhuisman:
I was in error when I deleted that old issue -- the availWidth/Height does not change when the browser window is resized.  This is a high-entropy detection and should (imo) be included in the fingerprint.

I also removed some trailing whitespace to get gulp to stop complaining.

### DIFF
--- a/fingerprint2.js
+++ b/fingerprint2.js
@@ -92,21 +92,32 @@
     },
     screenResolutionKey: function(keys) {
       if(!this.options.excludeScreenResolution) {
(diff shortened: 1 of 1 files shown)


Valve:
Thank you @mhuisman :heart:

Valve:
@mhuisman I've added you as a contributor to both FP and FP2, thanks for your code!

mhuisman:
Thanks @Valve!

PR #50, opened 2015-07-07
New canvas

antoinevastel:
I improved the way to use canvas by using techniques explained here : https://securehomes.esat.kuleuven.be/~gacar/persistent/the_web_never_forgets.pdf (section 3.2)

I also added a method to check if the user tampered with its browser. There are 2 errors when you run gulp but that is normal. It is used to check the consistencies of the browser.

Regards


mhuisman:
Why did you remove the canvas blending and winding?

fingerprint2.js
```diff
@@ -463,69 +470,25 @@
...
-      // https://www.browserleaks.com/canvas#how-does-it-work
-      var txt = "https://github.com/valve for PEACE in Ukraine!";
-      ctx.textBaseline = "top";
-      // ios8 specific font
-      ctx.font = "72px 'DamascusLight'";
+      // https://securehomes.esat.kuleuven.be/~gacar/persistent/the_web_never_forgets.pdf
+      ctx.textBaseline = "alphabetic";
       ctx.fillStyle = "#f60";
-      ctx.fillRect(2, 0, 1000, 70);
+      ctx.fillRect(125, 1, 62, 20);
```
Valve: did you verify the text fits into the rectangle?

Valve:
@antoinevastel , I currently fail to see how this will improve Canvas FP, could you please explain?

fingerprint2.js
```diff
@@ -463,69 +470,25 @@
...
-      // http://blogs.adobe.com/webplatform/2013/01/30/winding-rules-in-canvas/
-      // http://jsfiddle.net/NDYV8/19/
-      ctx.arc(75, 75, 75, 0, Math.PI * 2, true);
-      ctx.arc(75, 75, 25, 0, Math.PI * 2, true);
-      ctx.fill("evenodd");
-      result.push("canvas fp:" + canvas.toDataURL());
-      return result.join("~");
+      ctx.font = "18pt Arial";
+      ctx.fillText("Cwm fjordbank glyphs vext quiz, \ud83d\ude03", 4, 45);
+      return canvas.toDataURL();
```
mhuisman: +1 to the change to the fill text and do attempting to use system default fonts.  Both produce good entropy.

-1 to removing canvas blending and winding without explanation

antoinevastel: 1st the text fits in the rectangle. For the canvas blending and winding I removed it because they didn't talk about this in the article on canvas fingerprinting. However i can add it back.
Concerning why it should improve fingerprinting, look at the section 3.2 of this document : https://securehomes.esat.kuleuven.be/~gacar/persistent/the_web_never_forgets.pdf

antoinevastel:
I have added back canvas blending and winding.

Valve:
@antoinevastel I agree on all, thanks, it's just worth noting that the original paper you reference was written after FP1 was written, so they probably used some ideas from FP1 code, not the other way around.

FP1, in turn, took some ideas from browserleaks website.

Valve:
@antoinevastel I've made a minor change to keep eslint happy.

PR #52, opened 2015-07-08
add loader

ericorruption:


Valve:
It does not merge, you need to fetch and rebase off upstream

PR #57, opened 2015-07-16
Add detection for webgl support.

Schinkentanz:


fingerprint2.js
```diff
@@ -822,6 +822,16 @@
       var elem = document.createElement("canvas");
       return !!(elem.getContext && elem.getContext("2d"));
     },
+    isWebGlSupported: function() {
+      if (!this.isCanvasSupported() || !!window.WebGLRenderingContext) {
```
Valve: Is `!!window.WebGLRenderingContext` correct? It's double negation, returns true when supported.
I think this needs to test for positive, rather than negative, like if(this.isCanvas && !!window.WebGLRenderingContext) { true} else { false};

Schinkentanz:
Sry. That was a typo.. But i already tried it with Modernizr:
```javascript
new Fingerprint2({
  excludeWebGL: !Modernizr.webgl
}).get(function(fingerprintHash) {
  // do stuff
});
```
The problem is, that (e.g.) the samsung galaxy s2 seems to support webgl (```Modernizr.webgl returns true```) ... but when the webgl fingerprint is generated, an error occurs: ```undefined function getShaderPrecisionFormat```. Thats why added this special webgl check.

Valve:
Then it's a bug in the WebGL FP itself. Please use the Modernizr code to detect the availability of the WebGL and file a separate issue with getShaderPrecisionFormat. I have an old android device will test there soon.

Valve:
@Schinkentanz , thanks. Did you try running gulp?

fingerprint2.js
```diff
@@ -217,9 +217,19 @@
       return keys;
     },
     webglKey: function(keys) {
-      if(!this.options.excludeWebGL && this.isCanvasSupported()) {
+      var isWebGlExcluded = !!this.options.excludeWebGL,
+          isWebGlSupported = this.isWebGlSupported();
+
+      if (!isWebGlExcluded && isWebGlSupported) {
```
Valve: I think it's a mistake. You test for support of WebGL and then log that it's not supported.
Could you also please not introduce the isWebGlExcluded variable and operate solely on original options.excludeWebGL? I think it can be done w/out adding the variable.

fingerprint2.js
```diff
@@ -822,6 +832,24 @@
...
+      if (!this.isCanvasSupported() || !window.WebGLRenderingContext) {
+        return false;
+      }
+
+      var canvas = document.createElement("canvas"),
+          glContext = canvas.getContext && (canvas.getContext("webgl") || canvas.getContext("experimental-webgl"));
+
+      if (glContext && !glContext.getShaderPrecisionFormat) {
+        if (typeof NODEBUG === "undefined") {
+          this.log("This browser supports webgl, but lacks some functionality (getShaderPrecisionFormat not avaialable)");
```
Valve: I think we should still do WebGL FP, even if getShaderPrecisionFormat is not available. We should just skip this part in the WebGL code for those devices. Please add a separate issue for this, and remove this check.
I'll implement the fix for Galaxy S2 right now.

Valve:
Hey, @Schinkentanz , thanks for the code, if you fixed the code per my comments, I'd be happy to merge.

Schinkentanz:
Should be fine now :)

Valve:
Thanks!

Schinkentanz:
Np :) Could you please push a new version to npm? Thanks

Valve:
Done, I've tweaked your code a bit to add more logging, but the idea is the same. Thanks.

Schinkentanz:
:+1:

PR #67, opened 2015-08-06
Prevent older firefox versions (e.g. v25) to crash while checking for…

Schinkentanz:
… webgl support.

### DIFF
--- a/fingerprint2.js
+++ b/fingerprint2.js
@@ -864,8 +864,15 @@
       if (!this.isCanvasSupported()) {
         return false;
       }
+
       var canvas = document.createElement("canvas"),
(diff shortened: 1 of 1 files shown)


Valve:
Hey @Schinkentanz, thanks for the PR! I've been trying to figure out the issue with FF, but couldn't reliably reproduce it. Does it fail consistently for you?

Valve:
Looks like you've forgotten to run gulp. I'll fix it. I need to integrate gulp with Github somehow.

Schinkentanz:
Ups sry :)
This problem is consistently reproducible via browserstack.

PR #70, opened 2015-08-19
Performance: only call toLowerCase() once per variable

blx:
Tiny change to the `getHasLied` functions to only call `toLowerCase()` once per variable, instead of repeatedly.

I can add the minified gulp output if you'd like, I wasn't sure if you wanted to do that yourself.

Cheers,
Ben

### DIFF
--- a/fingerprint2.js
+++ b/fingerprint2.js
@@ -740,22 +740,22 @@
       return false;
     },
     getHasLiedOs: function(){
-      var userAgent = navigator.userAgent;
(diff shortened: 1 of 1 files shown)


Valve:
Out of the country now will merge in several days

Valve:
@blx , thanks, very nice readability update as well!

blx:
Glad to contribute! :)

Valve:
released new version

PR #81, opened 2015-10-22
In case of iPhone and iPad useragent, `getHasLiedOs()` always returns true.

bluele:
Even though gave a lowercase useragent, compared it with an uppercase string.

### DIFF
--- a/fingerprint2.js
+++ b/fingerprint2.js
@@ -760,7 +760,7 @@
         os = "Android";
       } else if(userAgent.indexOf("linux") >= 0){
         os = "Linux";
-      } else if(userAgent.indexOf("iPhone") >= 0 || userAgent.indexOf("iPad") >= 0 ){
+      } else if(userAgent.indexOf("iphone") >= 0 || userAgent.indexOf("ipad") >= 0 ){
         os = "iOS";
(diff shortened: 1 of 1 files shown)


Valve:
It is very strange that you have this, the correct names are indeed iPhone and iPad, not iphone and ipad, 
please see this: http://stackoverflow.com/a/2248420

Could you please post entire UserAgent strings for your iPad/iPhone?

bluele:
Of course, the correct names are iPhone and iPad. (ex. Mozilla/5.0(iPad; U; CPU iPhone OS 3_2 like Mac OS X; en-us) AppleWebKit/531.21.10 (KHTML, like Gecko) Version/4.0.4 Mobile/7B314 Safari/531.21.10)
But you called `navigator.userAgent.toLowerCase()` at this line: https://github.com/bluele/fingerprintjs2/blob/b4c4aa64e4d525ed0f94ae17f916e0ba454a2b03/fingerprint2.js#L750

Valve:
Oh, of course you're right, thanks! I actually didn't lowercase this, another contributor did in order to minimize the number of calls to 'toLowerCase'.

bluele:
Thanks!

PR #88, opened 2015-11-23
Return key => val array of data from which the hash is built

S-anasol:
In my case i want store all fingerprint data in DB.
But I have not found anything that can do it in your lib :)

So here my fast realisation.

Maybe you can do this idea better or "right-way".

or you can just merge my code :)

### DIFF
--- a/fingerprint2.js
+++ b/fingerprint2.js
@@ -1,5 +1,5 @@
 /*
-* Fingerprintjs2 1.0.0-rc2 - Modern & flexible browser fingerprint library v2
+* Fingerprintjs2 0.8.0 - Modern & flexible browser fingerprint library v2
 * https://github.com/Valve/fingerprintjs2
 * Copyright (c) 2015 Valentin Vasilyev (valentin.vasilyev@outlook.com)
 * Licensed under the MIT (http://www.opensource.org/licenses/mit-license.php) license.
@@ -57,7 +57,6 @@
     var defaultOptions = {
       swfContainerId: "fingerprintjs2",
       swfPath: "flash/compiled/FontList.swf",
-      detectScreenOrientation: true,
       sortPluginsFor: [/palemoon/i]
     };
     this.options = this.extend(options, defaultOptions);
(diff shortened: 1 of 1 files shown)


Valve:
This looks good to me, however you did some changes that I don't understand. Please see my comments.

fingerprint2.js
```diff
@@ -780,7 +784,7 @@
         os = "Android";
       } else if(userAgent.indexOf("linux") >= 0){
         os = "Linux";
-      } else if(userAgent.indexOf("iphone") >= 0 || userAgent.indexOf("ipad") >= 0 ){
```
Valve: Why making this change?

fingerprint2.js
```diff
@@ -492,7 +496,7 @@
     },
     touchSupportKey: function (keys) {
       if(!this.options.excludeTouchSupport){
-        keys.push(this.getTouchSupport());
+        keys.push({key: "getTouchSupport", value: this.getTouchSupport()});
```
Valve: these should have names without 'get' part, for example 'touch_support'.

fingerprint2.js
```diff
@@ -234,36 +238,36 @@
         }
         return keys;
       }
-      keys.push(this.getWebglFp());
+      keys.push({key: "getWebglFp", value: this.getWebglFp()});
```
Valve: try to use the short possible names, for example 'webgl' etc

fingerprint2.js
Valve: Please also make more descriptive commit messages. "change fingerprint.js" is a really bad commit message. It says nothing.

fingerprint2.js
S-anasol: i just used old version from composer, forgot to update it

fingerprint2.js
S-anasol: yeah i see.
I did Copy/Paste from right side :)

fingerprint2.js
S-anasol: Copy/Paste :smile:

S-anasol:
Should be better now :pray:

Valve:
I've merged, but please use 'contributing.md' next time. You obviously forgot to run gulp.

Valve:
Please add a separate PR with a spec, covering your new functionality. After that I will release a new version.

PR #89, opened 2015-11-24
update specs for checking returned components

S-anasol:
Added few tests.

Maybe we should check more? I have no idea about other checks :)

> SUCCESS: 15 specs, 0 failures, 0 skipped, 0 disabled in 11.034s.


specs/specs.js
```diff
@@ -89,5 +89,32 @@ describe("Fingerprint2", function () {
         });
       });
     });
+	
+	describe("return components", function () {
```
Valve: please fix indentation

specs/specs.js
```diff
@@ -89,5 +89,32 @@ describe("Fingerprint2", function () {
...
+      it("does it return components", function (done) {
```
Valve: should be imperative, e.g. it("returns components as a second argument to callback")

specs/specs.js
```diff
@@ -89,5 +89,32 @@ describe("Fingerprint2", function () {
...
+        var fp2 = new Fingerprint2();
+        fp2.get(function(result, components) {
+          expect(components).not.toBeNull();
+          done();
+        });
+      });
+	  
+      it("is first element user_agent", function (done) {
```
Valve: it ("returns user_agent as the first element"). i.e. the spec should read as a sentence

specs/specs.js
```diff
@@ -89,5 +89,32 @@ describe("Fingerprint2", function () {
...
+        var fp2 = new Fingerprint2();
+        fp2.get(function(result, components) {
+          expect(components[0].key).toEqual("user_agent");
+          done();
+        });
+      });
+	  
+      it("user_agent is PhantomJS", function (done) {
```
Valve: I would advise agains hardcoding PhantomJS, because we may not use it like tomorrow

Valve:
Looks good. Please add a check that the components is an array and that some of its individual elements are arrays too (fonts, plugins etc).

S-anasol:
How can i check that components is an array? :smile: 
Found only toBeArray in jasmine-expect, but have no idea how include it to tests.

Never used JS tests before :)

specs/specs.js
S-anasol: okay, removed browser test

specs/specs.js
S-anasol: done

specs/specs.js
S-anasol: done

specs/specs.js
S-anasol: done

S-anasol:
added jasmine-matchers for testing array/arrayOfObjects

specs/specs.js
```diff
@@ -89,5 +89,45 @@ describe("Fingerprint2", function () {
...
+          done();
+        });
+      });
+      
+      it("check if js_fonts component is array", function (done) {
+        var fp2 = new Fingerprint2();
+        fp2.get(function(result, components) {
+          for(var x = 0; x < components.length; x++) {
+            if(components[x].key == "js_fonts")
+            {
```
Valve: open curly on the same line please

specs/specs.js
```diff
@@ -89,5 +89,45 @@ describe("Fingerprint2", function () {
         });
       });
     });
+    
+    describe("returns components as a second argument to callback", function () {
+      it("does it return components", function (done) {
```
Valve: it("returns components")

specs/specs.js
```diff
@@ -89,5 +89,45 @@ describe("Fingerprint2", function () {
...
+        var fp2 = new Fingerprint2();
+        fp2.get(function(result, components) {
+          expect(components).not.toBeNull();
+          done();
+        });
+      });
+      
+      it("check if returned components is array", function (done) {
```
Valve: it("checks if..

specs/specs.js
```diff
@@ -89,5 +89,45 @@ describe("Fingerprint2", function () {
...
+        var fp2 = new Fingerprint2();
+        fp2.get(function(result, components) {
+          expect(components).toBeArrayOfObjects();
+          done();
+        });
+      });
+      
+      it("check if js_fonts component is array", function (done) {
```
Valve: check => checks

S-anasol:
done :+1:

Valve:
thanks

PR #115, opened 2016-01-15
New option to exclude available resolution separately

nuschk:


fingerprint2.js
```diff
@@ -144,11 +144,13 @@
       if(!this.options.excludeScreenResolution) {
         return this.getScreenResolution(keys);
       }
+      if (!this.options.excludeAvailableResolution) {
```
Valve: please call this excludeAvailableScreenResolution to be consistent

Valve:
@nuschk did you get a chance to finalize this PR?

nuschk:
@Valve I've changed the naming as you suggested. I think the PR is fine. Although, I must admit, I haven't really dug into the code too much. I didn't find anything to add to the documentation nor any tests, I hope that's correct.

Valve:
@nuschk thanks for the update. Please add a spec to cover this (similar to this: https://github.com/Valve/fingerprintjs2/blob/master/specs/specs.js#L74) and we should be good to go

nuschk:
@Valve Oops, I missed that. Will do. Thanks!

nuschk:
@Valve So, I've added the spec and also made the available screen res independent from the "normal" screen res. I hope that somehow fits the architecture you had in mind.

Valve:
thanks, released v.1.1.0

nuschk:
Thanks! :+1:

Issue #126, opened 2016-02-22
Font detection slow


msp:
Hey, thanks for the library!

Any idea why the font detection is slow? My numbers seem way above from what you quote in the README. I'm using the default config.

Approx 4 seconds in Chrome and longer in Safari..

Fingerprinting INIT took 0.5349999999998545 milliseconds.

application.js:75591 userAgentKey:0.45499999999992724 milliseconds.
application.js:75595 languageKey:0.1449999999999818 milliseconds.
application.js:75599 colorDepthKey:0.14000000000032742 milliseconds.
application.js:75603 screenResolutionKey:0.3049999999998363 milliseconds.
application.js:75607 availableScreenResolutionKey:0.2899999999999636 milliseconds.
application.js:75611 timezoneOffsetKey:0.16499999999996362 milliseconds.
application.js:75615 sessionStorageKey:0.12500000000045475 milliseconds.
application.js:75619 localStorageKey:0.26499999999987267 milliseconds.
application.js:75623 indexedDbKey:0.1999999999998181 milliseconds.
application.js:75627 addBehaviorKey:0.205000000000382 milliseconds.
application.js:75631 openDatabaseKey:0.17000000000007276 milliseconds.
application.js:75635 cpuClassKey:0.525000000000091 milliseconds.
application.js:75639 platformKey:0.38499999999976353 milliseconds.
application.js:75643 doNotTrackKey:0.15000000000009095 milliseconds.
application.js:75647 pluginsKey:1.1000000000003638 milliseconds.
application.js:75651 canvasKey:23.725000000000364 milliseconds.
application.js:75655 webglKey:82.53499999999985 milliseconds.
application.js:75659 adBlockKey:0.2250000000003638 milliseconds.
application.js:75664 hasLiedLanguagesKey:0.15000000000009095 milliseconds.
application.js:75668 hasLiedResolutionKey:0.14000000000032742 milliseconds.
application.js:75672 hasLiedOsKey:0.3599999999996726 milliseconds.
application.js:75676 hasLiedBrowserKey:0.5149999999994179 milliseconds.
application.js:75680 touchSupportKey:0.3700000000003456 milliseconds.

application.js:76013 FONTS:4081.925 milliseconds.
application.js:75694 fontsKey:4172.885 milliseconds.

XX:585 Fingerprinting GET took 4292.765 milliseconds.

Valve:
[...]

PR #128, opened 2016-02-23
Font iteration 'bug'

msp:
See #126


fingerprint2.js
```diff
@@ -359,13 +359,13 @@
         }
         var detect = function (font) {
             var detected = false;
-            for (var index in baseFonts) {
+            baseFonts.forEach(function() {
```
Valve: AFAIK, this will not work in IE8

msp: Looks like you're right:

http://caniuse.com/#search=forEach

I can change to a C style for loop..

Valve: Thanks!

JustinTArthur:
:+1: for this. For the sake of speed on older browsers, I'd recommend storing the length at loop initialization as is done here
https://github.com/Valve/fingerprintjs2/blob/62b276871fc67c9e58c8bdea03cc830ed8105c7c/fingerprint2.js#L422

Valve:
Thanks, I will review and merge tomorrow morning!

Valve:
Thanks, released 1.1.1

PR #129, opened 2016-02-27
Dev fixes #103

### LINKED ISSUE #103: canvas fingerprint is different for different zoom levels on Firefox
When I zoom the page in FF, I get different FPs, because the  `canvas.toDataURL` returns different results for different zoom levels.

RobertBorg:
## Changes
* fixed gulp-eslint error by changing false to 0 in .eslintrc
* implemented lookup assisted determination of scaleResolution via https://en.wikipedia.org/wiki/List_of_common_resolutions#Computer_graphics using difference in aspect ratio as tie breaker if multiple suitable candidates are found and falling back to old method if no candidates are found. It's not perfect.
* changed CanvasFP generation for Firefox to use px instead of pt in fonts. This fixes #103 .


Valve:
Could you please split this PR into multiple? I cannot merge this into master now in its entirety, but I would like to test and merge px vs pt in firefox.

Issue #13, opened 2015-03-26
Return Promise


Valve:
If native ES6 promise available.
Open questions:

1. What should happen if any polyfill is available.
2. If the fingerprint returning method will be ok to implement as promise-returning, callback-accepting to handle both presence and absence of the promise impl.

Valve:
v2 will stop using Flash integration and will become synchronous as a result, so no promise is necessary.

PR #131, opened 2016-03-10
Media Devices Integration

coltmeister:
Hello,

I've done a small refactor for asynchronous functions in this library, and also added a media devices integration. This integration can be disabled by setting the "enableMediaDevices" to false in the options object. Enabling the 'media devices' key will show greater accuracy when comparing fingerprints of the same browser over short periods of time. However, for a particular browser fingerprint, if a media device is disconnected, that browser fingerprint will change; thus, for **less accuracy** over a shorter period but **more accuracy** overall, 'media devices' should be disabled. One example of a useful case of enabling 'media devices' would be tracking a user (who has cookie's disabled) from a company's landing page to that company's home page, since the actions of recording the fingerprint (visiting the landing page) and "looking-up" the fingerprint (visiting the home page) will be closely related temporally. An example of a _non-useful_ case of enabling 'media devices' would be storing information about an anonymous visitor for the purpose of "remembering" the visitor over a given amount of time. 

I have tried my best to code this in the same style that it was written! I ran the jasmine tests, they all passed. I also verified the consistency of my 'browser fingerprint' locally (with mozilla firefox, chrome, and safari).


Valve:
Thanks! I will, however, be able to look into it on weekend only. Sorry for the delay.

coltmeister:
No problem, there's no need to rush. I will go ahead and say, I shouldn't
have included the npm log in this pull request, that was my mistake. Also
this is the first time I've ever contributed to an open source project so,
if you have time while you're looking over, please feel free to criticize
my code!

Valve:
@coltmeister haha, no problem. Most people never contribute, but you did and it's a very good thing!

Valve:
I would, however, ask you to run the gulp, if you haven't already, to produce the minified version of the script and to catch potential stylistic errors and to squash all your commits into one with a good commit message, explaining what you have added and how it affects the FPing :)

coltmeister:
It was pretty fun, I'm planning on contributing more (it's also a really
good learning exercise)!
I ran the gulp file when I thought I was finished, then the tests failed
and I refactored some code, and I forgot to run the gulp again. This is a
noobie question but, when I run the gulp, how do I cancel my pull request
so that I can send another with the updated changes?
Thanks!

Valve:
You don't need to cancel your PR, just add more commits to your fork and the PR will be updated correspondingly

coltmeister:
Alright awesome, just took care of that. Thanks, have a great week!

Valve:
@coltmeister please squash git commits into one with a good commit message, explaining the essense of the commit code

fingerprint2.js
```diff
@@ -428,6 +460,23 @@
         done(keys);
       }, 1);
     },
+    mediaDevicesKey: function(keys, done) {
+        //Adds the device key for each media device to the devicesArr
+        if(navigator.mediaDevices && !this.options.excludeMediaDevices){
```
Valve: option check should come first, because we don't want to touch mediaDevices if user has them excluded

fingerprint2.js
```diff
@@ -428,6 +460,23 @@
...
+            var devicesArr = [];
```
Valve: Please don't use Arr suffix for arrays, plural nouns are sufficient

fingerprint2.js
```diff
@@ -428,6 +460,23 @@
...
+            var that = this;
+            navigator.mediaDevices.enumerateDevices().then(function(devices){
```
Valve: Media devices are only supported in Chrome 45+ with a flag, so in reality they are not supported currently in all major browsers, please see this: https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices

Please add error handling for cases when it's not supported. I believe current code will raise errors in browsers that don't support them.

You're also using promise here, so this ticket should be blocked by #13

fingerprint2.js
```diff
@@ -104,19 +107,48 @@
...
-        return done(murmur, newKeys);
-      });
+
+      //Asynchronous keys, must be the case sensitive name of the asynchronous function
+      //(this is done to protect context integrity)
+
+      var asyncKeyFunctions = ["fontsKey", "mediaDevicesKey"];
+      this.addAsynchronousKeys(keys, asyncKeyFunctions, done);
+    },
+    addAsynchronousKeys: function(currentKeys, asyncFunctions, done) {
```
Valve: I think this should be done after #13 is done and should be implemented as a promise chain internally

Valve:
I would also ask you to create an issue and link this PR to the issue. Current support for media devices is so scarce that I doubt it will benefit the library in the web that we have in 2016, so I'll probably mark it for v3 milestone

coltmeister:
Thanks for looking it over, I'll look into #13!

PR #134, opened 2016-03-18
Adding css to the span element to avoid its display.

skapoor:
For some browsers like iOS9, This span shows up for some micro Sec and then gets hidden which creates a bad user experience. This css will make sure that this element does not show up in the screen.


Valve:
Thanks for the PR!

Please read contributing.md :) You forgot to run gulp

skapoor:
Thanks @Valve 

Ran `gulp` and fixed the issues as well.

fingerprint2.js
```diff
@@ -345,6 +345,13 @@
 
         // create a SPAN in the document to get the width of the text we use to test
         var s = document.createElement("span");
+        /*
+         * We need this css as in some weird browser this
+         * span elements shows up for a microSec which creates a
+         * bad user experience
+         */
+        s.style.position = "absolute";
+        s.style.left = "-9999px";
```
mhuisman: Could we not display:none here instead?

Valve:
@mhuisman it will not work: https://jsbin.com/pedukutili/edit?html,output

@skapoor thanks for you PR!

Did the detected font array was identical when detected with old method?
In JsBin I see that OffsetHeight is slightly different when using your method.

Valve:
Tested on my end, looks fine. Released v1.1.2

PR #139, opened 2016-03-22
Added support for synchronous calculation. Fixes #138

### LINKED ISSUE #138: Feature Request : Synchronous Calculation
In one of our projects we need to calculate fingerprint synchronous. We patched the library as of now.

It's a nice to have based on a flag in options.

sasidhar:


Valve:
I cannot merge it into master, sorry, current master must stay async. I will be able to change this in v2 where there will be no Flash integration.

Issue #142, opened 2016-04-05
Plugin detection in at least IE 11 causing prompt for users


StefanWallin:
Users with IE 11 on win 7(only verified browser) and quicktime installed on their machines receive the following permission popup in their browse: 
`This webpage wants to run the following add-on: 'Apple' from 'Apple Inc.'`
<img width="969" alt="This webpage wants to run the following add-on: 'Apple' from 'Apple Inc.'" src="https://cloud.githubusercontent.com/assets/457653/14282139/5d92e726-fb3d-11e5-9b05-06f43ba595c4.png">

Commenting out the detection for `QuickTimeCheckObject.QuickTimeCheck.1` alleviates this issue. 

**Questions are:**

1. Is this intended behaviour? 
2. Does this behaviour also exist for other plugins in IE?

**Repliaction:**
- Install win7IE11 image from modern.ie
- Install quicktime from apple.com
- visit site with active fingerprint2js

(internal ping: @triptec)

Valve:
@StefanWallin oh, certainly not intended, it's a bug.
Unfortunately, IE is a second-class citizen when it comes to my testing, because I'm either on Linux or Mac, so cannot reproduce it now. Please comment out the offending lines for now

triptec:
I wonder if this isn't true for all the plugins for IE 11, I'd like a option to disable the plugin detection on IE, I believe (not confirmed) that the popup has to do with the `new ActiveXObject(name);`

Valve:
Oh my, I definitely need a Windows box to test all this stuff

StefanWallin:
You can use a modern.ie virtualbox image to replicate on your mac/linux, that's what I used. But I had to of course install quicktime within that VM…

Valve:
Fixed in #144

PR #144, opened 2016-04-22
Add excludeIEPlugins option

triptec:
In at least newer versions of Internet Explorer the plugin detection triggers a user confirmation bar as seen in #142.
I would like a option to make sure that doesn't happen.


Valve:
👍

PR #162, opened 2016-06-07
Clean up body after checking

ugwe43to874nf4:
It's better to remove the inserted node, because it may mess up the page layout.


Valve:
Hi, thanks for contributing, please read contributing.md

PR #163, opened 2016-06-10
Add detection of pixel ratio

biphobe:


Valve:
Is it pixel density or pixel ratio?
These are different things. I think pixel ratio is a very simple thing, common across many devices, while pixel density can be very specific per model.  (like 334 DPI)

biphobe:
@Valve You're right, I have added feature request to check pixel density back to README.md, I will think about doing this some other time.

I ask you to merge this change nonetheless since this check is extremely cheap to do and, after all, pixel ratio can vary across devices with similar configurations.

Valve:
Sure, I'll merge it over the weekend

PR #187, opened 2016-09-24
Skip browser version

S-anasol:
Skip browser version, to avoid hash changing after browser update.
Support only for most common browsers and phantomjs for testing.

I have seen in gitter chat that such function would be useful, long time ago.
Now i want to use it by myself, but it still not available, so i decide to code it :)

New option **excludeBrowserVersion** will remove browser version from useragent string.

But not sure about version of other useragent components like _AppleWebKit/537.36, Safari/537.36_.
Will it change with browser update? Or just legacy versions of browser engine.


Valve:
Thank you for your work, @S-anasol, but I feel that I still want to add the callback route to use some established UA parsers. Your solution will quickly become outdated and unusable, I believe we should rely on external parsers to do this job. I will close this PR.

I think a fingerprint function should accept a special option with callback function where you can parse the UA, like follows:

```javascript
var options = {
  preprocess: function(key, value) {
     if(key === "userAgent") {
        var parsedUserAgent = SomeUserAgentParser.parse(value);
        var userAgentMinusVersion = parsedUserAgent.base_platform + ' ' + parsedUserAgent.browser_name;
        return userAGentMinusVersion;
   });

var fp = new Fingerprint(options);
```

PR #188, opened 2016-09-28
fix js error if pair.value is undefined

Handsome2734:
fix js error if pair.value is undefined, which blocks generating fingerprint.

### DIFF
--- a/fingerprint2.js
+++ b/fingerprint2.js
@@ -111,7 +111,7 @@
         var values = [];
         that.each(newKeys, function(pair) {
           var value = pair.value;
-          if (typeof pair.value.join !== "undefined") {
+          if (pair.value && typeof pair.value.join !== "undefined") {
             value = pair.value.join(";");
           }
           values.push(value);


Valve:
👍 thanks, will review and merge over the weekend

fingerprint2.js
```diff
@@ -111,7 +111,7 @@
         var values = [];
         that.each(newKeys, function(pair) {
           var value = pair.value;
-          if (typeof pair.value.join !== "undefined") {
+          if (pair.value && typeof pair.value.join !== "undefined") {
```
Valve: Please let me know in what cases pair can be null? That does not fix the root cause of the error but only hides it

Handsome2734: http://www.youku.com/
Visit this site in mobile emulation mode in Chrome, and screen object will be a String rather than an Object, which is set in html inline script.
So every `pair.value` get from `screen` will be undefined.

Valve:
I cannot reproduce it, sorry, I'm using the latest chrome + iPhone6 emulation

elad-yosifon:
Thanks for supplying this reproducer.. I was able to reproduce it on my end.
The issue is in the ```colorDepthKey``` method.

According to MDN (https://developer.mozilla.org/en-US/docs/Web/API/Screen/pixelDepth), this value should always return 24. I'm assuming that this feature was added in order to identify devices which violate this spec.

@Valve - Is my assumption correct?

Valve:
Yes, but it's not a violation, but rather detection.

elad-yosifon:
if this is indeed the case, I think that in case of a falsy ```screen.pixelDepth```, we should return 0 or -1 
(an intentional invalid value, that "detects" devices that violate this spec).

Valve:
yeah, ok

elad-yosifon:
@Valve - do you want to add this fixup to v2 or solve it in master?

Valve:
Let's fix it in master

elad-yosifon:
pushed a fixup to master.. closing this pull request.

Valve:
thanks @elad-yosifon

PR #192, opened 2016-10-02
Remove gulp dependency

felquis:
Now the steps to run the program is:

```
npm install
npm run build
```

`npm run build` will lint the code with eslint and uglify using uglify-js, the same as before but now with less dependencies


felquis:
I only did it cause this script is really small, it is so simple that could be done the same without a dependency like gulp.

Valve:
I'm not sure about this change, @felquis . Gulp is a big dependency but a dependency almost any project has. It adds some overhead but the gulpfile is easier to read than your custom script, imo.

felquis:
You are right! I just wanted to give it a try :)

Valve:
thanks @felquis !

Issue #185, opened 2016-09-21
Two separate websites, same library version, same browser version, different fingerprint.


bazaarcodinglabs:
I have two separate websites, website A and website B, each using the same version of the fingerprint2 library.  Using the same browser, I have noticed that the fingerprint was different on website A.  It is different because one font ("Wingdings 3") was missing in the 'js_fonts' array in website A.  The font was not being detected because on website A, the "line-height" on website A was explicitly set to 14px for all divs - this is being done on our internal stylesheet.  

I found the solution by resetting the "line-height" on the span to "normal", this allows the "line-height" to be the actual height of a particular font, rather than allowing it to inherit the line-height from another style.

The solution is as follows, its a one liner (line 12), could you consider including it in your library please?

```

1.                 // creates a span where the fonts will be loaded
2.                 var createSpan = function() {
3.                     var s = document.createElement("span");
4.                     /*
5.                      * We need this css as in some weird browser this
6.                      * span elements shows up for a microSec which creates a
7.                      * bad user experience
8.                      */
9.                      s.style.position = "absolute";
10.                     s.style.left = "-9999px";
11.                     s.style.fontSize = testSize;
12.                     s.style.lineHeight = "normal";
13.                     s.innerHTML = testString;
14.                     return s;
15.                 };

```

Many thanks.

Barry Edwards

Valve:
Yes, of course, please do a PR and I'll merge it

bazaarcodinglabs:
Sure, will do tomorrow.  Thank you.

Valve:
Fixed in #193

PR #193, opened 2016-10-04
Fix for issue #185. Set line height for proper font detection

imjustabill:
Implementing the fix from https://github.com/Valve/fingerprintjs2/issues/185


Valve:
:+1:

Issue #186, opened 2016-09-22
Two separate websites, same library version, same browser version, different fingerprint.


bazaarcodinglabs:
I have come across another issue regarding the canvas.  I have two separate websites, website A and website B, each using the same version of the fingerprint2 library. Using the same browser, I have noticed that the fingerprint was different on website A.  When I pass in 'excludeCanvas: true' into the Fingerprint2 constructor for both websites, the fingerprints are the same.  Any ideas why this might be happening please?

bazaarcodinglabs:
I've noticed that when I open index.html in a browser and use the minified version of fingerprint2.js, it produces a different fingerprint than using the source file.  Any ideas please?

titel:
I'm having the same issue (I've tested on multiple domains). It's like if the hostname of the page is taken into account if when generating the hash. Within the same domain the hash is consistent however.

I'm willing to provide a live example in private. Please contact me if required.

bazaarcodinglabs:
I've figured it out.  There are a couple of reasons.
1.  One reason is that the line height on a web page can affect the line
height of the font as suggested in my first issue.
2.  The other issue is, when either grunt or gulp minifies (uglifies) the
file, the unicode characters get converted to their actual characters which
affects the fingerprint.  To fix this in your code base, simply do this in
your gulp file: uglify({output: {ascii_only:true}}).  See
https://github.com/terinjokes/gulp-uglify/issues/120 for more details.

Let me know if you have any further questions.

Cheers,

Barry

On Fri, Sep 30, 2016 at 4:35 PM, Constantin Tovisi <notifications@github.com

Valve:
Thank you!

bazaarcodinglabs:
No problem, glad I could help!

On Fri, Sep 30, 2016 at 9:25 PM, Valentin Vasilyev <notifications@github.com

Valve:
Fixed in #194

PR #194, opened 2016-10-05
Set uglify output to ascii only. Fix for issue #186

imjustabill:
Implemented fix from https://github.com/Valve/fingerprintjs2/issues/186


Valve:
Thanks 👍

PR #197, opened 2016-10-05
Resolved DoNotTrack Issues with IE 9-10, IE 11, and Safari 7.1.3+

motianinikhil:
Different browsers use different fields that indicate the user's do-not-track setting. This PR makes the library's doNotTrack function cover IE 9-10, IE 11, and Safari 7.1.3+ usecases.

For more info, Check footnotes on MDN Page - https://developer.mozilla.org/en-US/docs/Web/API/Navigator/doNotTrack


Valve:
Thanks 👍

Issue #189, opened 2016-09-28
IndexedDB: A mutation operation was attempted on a database that did not allow mutations.


jonashaag:
This just bubbled up in our Javascript error tracking:

```
A mutation operation was attempted on a database that did not allow mutations.
  at t.prototype.hasIndexedDB
  at t.prototype.indexedDbKey
  at t.prototype.get
```

Firefox 47, Windows 7

Valve:
Thanks for letting me know. Please do a PR if this fix is important to you. I'll try to fix it over the weekend.

jonashaag:
This has been fixed already

PR #206, opened 2016-11-08
￼￼hasIndexedDB fails within iframe when third party cookies are off

leafo:
Similar to the other `try`/`catch` statements used for checking storage, we need to wrap `hasIndexedDB`.

When third party cookies are disabled and fingerprintjs runs in an iframe, trying to access `indexedDB` would result in an exception preventing the fingerprint from bring calculated.

### DIFF
--- a/fingerprint2.js
+++ b/fingerprint2.js
@@ -626,7 +626,11 @@
       }
     },
     hasIndexedDB: function (){
-      return !!window.indexedDB;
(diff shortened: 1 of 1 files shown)


bluestreek18:
This also happens in ff 47 incognito as it won't allow access. #189

Valve:
:+1:

PR #212, opened 2016-12-05
Preprocess components (strip browser version etc...)

S-anasol:
Second try to do this #187  😄


Valve:
@S-anasol sorry for the delay, I will review this on the weekend :+1:

S-anasol:
any weekend on any year 😆

Valve:
patience, my friend :) you can use your fork for a time being

S-anasol:
just joking 😹

S-anasol:
Any news? updated to latest master.

drstyv:
Any news on this or any of the other updates that are pending?

jonashaag:
@Valve this patch looks OK to me -- not exactly the best code quality but correctly implemented.

S-anasol:
@drstyv @jonashaag 
https://github.com/Valve/fingerprintjs2/commit/be2af5358a82e0d103cf088705dfe75788003124

FP is not maintained now probably

Valve:
@S-anasol I no longer maintain the lib, but  @jonashaag is taking on the role of a maintainer :)
If the code quality leaves to be desires, please point out the bad parts. I would generally not accept bad quality PRs.

@S-anasol please work with @jonashaag to improve code quality and he will merge it :)

README.md
```diff
@@ -110,6 +110,25 @@ new Fingerprint2({
 });
 ```
 
+#### `preprocessor` option
+
+Since some components can be changed frequently, this option may help you improve hash accuracy. You can modify any component value before hash calculation. For example: strip browser version from user agent.
```
jonashaag: This section should start with an explanation of what the `preprocessor` option does, something along the lines of "Function that is called with each component value that may be used to modify component values before computing the fingerprint". The user agent example is good.

README.md
```diff
@@ -123,15 +142,15 @@ To start a web server you can try using one of the following:
 
 * Python 2.x
 
-	`python -m SimpleHTTPServer 8080`
+    `python -m SimpleHTTPServer 8080`
```
jonashaag: These seem like unrelated changes

fingerprint2.js
```diff
@@ -52,7 +52,18 @@
       return target;
     },
     get: function(done){
-      var keys = [];
+      var that = this;
+      var keys = {
+          data: [],
+          push: function(pair){
```
jonashaag: I'd rather see this as an extra method/function with a meaningful name, something like `addPreprocessedComponent` or `preprocessAndAdd`. Also `tmpX` aren't good variable names at all, please use something more meaningful here.

I also don't like that you created a "looks-like-an-array-but-isn't" object here -- either it looks like an array, then it should be an array too, or make it look completely different. Otherwise people will get confused.

specs/specs.js
```diff
@@ -243,5 +243,40 @@ describe("Fingerprint2", function () {
         });
       });
     });
+
+    describe("preprocessor", function () {
+      it("checks if user_agent not processed by default", function (done) {
```
jonashaag: This description should not mention user agent. It should mention preprocessor instead. You're not testing something specific to the user agent field. You're testing the preprocessor option.

specs/specs.js
```diff
@@ -243,5 +243,40 @@ describe("Fingerprint2", function () {
...
+          for(var x = 0; x < components.length; x++) {
+            if(components[x].key == "user_agent") {
+                expect(components[x].value).not.toEqual("MyUserAgent");
+            }
+          }
+          done();
+        });
+      });
+      
+      it("checks if user_agent processed with user function", function (done) {
```
jonashaag: same here

fingerprint2.js
S-anasol: > looks-like-an-array-but-isn't
just to avoid many lines changes as it now :)

README.md
```diff
@@ -110,6 +110,26 @@ new Fingerprint2({
...
+Function that is called with each component value that may be used to modify component values before computing the fingerprint.
+For example: strip browser version from user agent.
+
+```
+new Fingerprint2({
+  preprocessor: function(key, value){
+    if(key == "user_agent"){
+      var parsedUserAgent = SomeUserAgentParser.parse(value);
+      var userAgentMinusVersion = parsedUserAgent.base_platform + ' ' + parsedUserAgent.browser_name;
+      return userAgentMinusVersion;
```
jonashaag: I think the example is a bit hard to understand. Could we either

- simplify the example, maybe using regex substitution for the version number (not sure if such regex exists?)
- or use a concrete user agent library (just the real API, no imports etc.)

fingerprint2.js
```diff
@@ -52,7 +52,18 @@
       return target;
     },
     get: function(done){
-      var keys = [];
+      var that = this;
+      var keys = {
+          data: [],
+          addPreprocessedComponent: function(pair){
+              var componentKey = pair.key;
```
jonashaag: IIUC this alias isn't necessary, we can simply use `pair.key` everywhere, is that correct?

jonashaag:
Awesome! Thanks a lot!

drstyv:
Awesome to see new updates again!

Valve:
@dangayle thanks to @jonashaag; he was very kind to take the project under his wing. Thank you Jonas!

PR #214, opened 2016-12-15
Update README.md

gaborgulyas:
Error in the description: instead of extendedFontList you have to use extendedJsFonts to enable the extended javascript font detection.


Valve:
:+1:

PR #230, opened 2017-02-06
main section shouldn't include minified version

Lapkritinis:
bower.json main session shouldn't include minified version of js.
https://github.com/bower/spec/blob/master/json.md#main


Valve:
then we should modify the bower file to include the minified version too?

Lapkritinis:
No, only "normal" script. I see other projects having two versions in "dist" folder - full and minified. But in main section there is only full.

PR #244, opened 2017-03-30
delete console logs in non minified version

jacek213:


Valve:
thanks, I'll release a new version today

PR #255, opened 2017-06-11
patch

wormen:


jonashaag:
Hi! Aren't these both subsets of the user agent component, and thus already included in the current fingerprint components?

jonashaag:
Closing due to inactivity.

PR #272, opened 2017-10-28
Add detection for Worker API support

jenic:
Small addition to add [Worker API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers) detection support. Not sure if useful...

### DIFF
--- a/fingerprint2.js
+++ b/fingerprint2.js
@@ -86,6 +86,7 @@
       keys = this.hasLiedBrowserKey(keys)
       keys = this.touchSupportKey(keys)
       keys = this.customEntropyFunction(keys)
+      keys = this.workerApiKey(keys)
       this.fontsKey(keys, function (newKeys) {
         var values = []
         that.each(newKeys.data, function (pair) {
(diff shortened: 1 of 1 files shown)


jonashaag:
Hm. Does this add any entropy over the user agent string? Can web workers be disabled by people? Are they disabled automatically in incognito? Does this somehow depend on your hardware setup?

jenic:
Very good questions I believe in firefox they can be disabled with `dom.workers.enabled` and `dom.serviceworkers.enabled` please correct if wrong. I don't believe they are disabled in incognito. I don't currently have a lot of time to test things, sorry bout that.

jonashaag:
Are you using this in a project and does it help you with fingerprinting your users or did you just think this could be a neat addition to the components? Like, do you have a use case?

jenic:
I'm not currently using this, no. I have plans to use it along with some mobile features coming down the pipeline in future browser releases. I use this project itself for short-term servers of various operating lengths and my specific use case calls for as much identifying information as possible.

I myself am skeptical as to its usefulness and only sent it to you on the off chance you might want it.

jonashaag:
OK, closing this then. If someone thinks this should be added to fingerprintjs, feel free to comment here.

PR #278, opened 2017-11-18
Add JSDoc annotations to fully support closure-compiler

ChadKillingsworth:
I've been using and enjoying the library. However with closure-compiler, I have to maintain a custom build to add a few JSDoc annotations.

These changes make the library fully compatible with Closure-Compiler ADVANCED mode.

### DIFF
--- a/fingerprint2.js
+++ b/fingerprint2.js
@@ -21,6 +21,10 @@
   if (typeof window.define === 'function' && window.define.amd) { window.define(definition) } else if (typeof module !== 'undefined' && module.exports) { module.exports = definition() } else if (context.exports) { context.exports = definition() } else { context[name] = definition() }
 })('Fingerprint2', this, function () {
   'use strict'
+  /**
+   * @constructor
+   * @param {Object=} options
+   */
   var Fingerprint2 = function (options) {
     if (!(this instanceof Fingerprint2)) {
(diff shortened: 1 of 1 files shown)


jonashaag:
Thanks!

How can we make sure Closure Compiler annotations aren't going to break in future releases? Is there some linter/checker that we can integrate into CI?

fingerprint2.js
```diff
@@ -56,7 +60,7 @@
           if (typeof that.options.preprocessor === 'function') {
             componentValue = that.options.preprocessor(pair.key, componentValue)
           }
-          this.data.push({key: pair.key, value: componentValue})
+          keys.data.push({key: pair.key, value: componentValue})
```
jonashaag: Ouch, this must have been a merge error. Thanks for fixing this

ChadKillingsworth:
> How can we make sure Closure Compiler annotations aren't going to break in future releases? Is there some linter/checker that we can integrate into CI?

If you are interested, I can PR a small test case that just runs the library through the compiler with all the checks enabled and ensures there are no warnings or errors.

fingerprint2.js
ChadKillingsworth: Closure-compiler complained loudly about that one.

jonashaag:
That would be awesome!

PR #279, opened 2017-11-29
Fixed some bugs in some environments.

hiuny:


jonashaag:
Thanks for the patches. For the new option we'll require some documentation in readme, wiki, and a test.

Could you go until detail what the problem with the "strange characters" is? How can this crash encodeUriComponent?

hiuny:
```javascript
// code for example
new Fingerprint2().get((hash, components) => {
  encodeURIComponent(JSON.stringify(components)); // error occurred. "URIError: malformed URI sequence"
});
```

I collected the error log.
There were several kinds of errors,
All of these errors occurred in Internet Explorer 11.
And there were strange characters.

```
1. "SniperPlayerV2::SniperPlayerV2/1.2.3.1::application/x-sniperplayer-v2~界ᆈv?༙茁ᆎ"
  This characters > 界ᆈv?༙茁ᆎ

2. "AdobeAAMDetect::Creative Cloud Desktop Plugin.v_3_0_0_0::application/x-adobeaamdetect~ນ?ນ?ນ"
  This characters > ນ?ນ?ນ
```

However, putting the string as an `encodeURIComponent()` parameter does not cause an error.

```javascript
encodeURIComponent("界ᆈv?༙茁ᆎ"); // "%E7%95%8C%E1%86%88v%3F%E0%BC%99%E8%8C%81%E1%86%8E"
encodeURIComponent("ນ?ນ?ນ"); // "%E0%BA%99%3F%E0%BA%99%3F%E0%BA%99"
```

It is inferred that the character causing the error is lost during network transmission.
I will reflect in the service, `excludeIERegularPlugins: true`.
So I'm not sure if this will fix the error yet.
But I think it's okay to add this option. :-)

jonashaag:
Interesting. I committed the rest of the changes, but I'd like to wait with the `excludeIERegularPlugins` option until 1) you have more data from your patched version and 2) we have documentation for the option.

hiuny:
ok~~ thanks~~

jonashaag:
Closing due to inactivity

PR #291, opened 2018-01-09
Introduced Promise API

### LINKED ISSUE #287: Poor async error handling
I am trying to integrate fingerprint2 in "safe" way to avoid any JS problems on client side.
The idea is to try compute FP using default (empty) options and in case of an error, modify options (eg. set `excludeWebGL` to `true`) and try to compute again.
I also have to migrate it into `Promise` to be compatible with the rest of my project.

Unfortunately the simplest solution
```
var fingerprint = new Promise(function(resolve, reject) {
  try {
    new Fingerprint2(options).get(resolve);
  } catch (e) {
    reject(e);
  }
});

fingerprint.then(alert, alert);
```
won't always work due to your `setTimeout` usage (https://github.com/Valve/fingerprintjs2/blob/master/fingerprint2.js#L323). If the error occurs inside the `setTimeout` callback function, the JS just errors into console and the `done` callback is never called.

# Steps to reproduce:

1. Add code `pair = null` just before https://github.com/Valve/fingerprintjs2/blob/master/fingerprint2.js#L97 (to simulate NPE)
```
pair = null
var value = pair.value
```
2. Watch the library do nothing with your callback and also not propagating the error up. Only printing error to console.
```
try {
  new Fingerprint2({}).get(alert);
} catch (e) {
  alert(e);
}
```

I have found this issue thanks to FF 52.5.3 (ESR). It blocks fingerprint in it's default configuration and gives these 2 errors:

- Error: WebGL: Disallowing antialiased backbuffers due to blacklisting.
- TypeError: pair is null - fingerprint2.js:97

# Solution
Simply add the `failed` callback along with the `done` callback. Enclose the inside of `get` function into `try-catch` block and call the `failed` callback in case of an error.
Also propagate the `failed` callback into the `fontsKey` and `jsFontsKey` functions. In the `jsFontsKey` function use `try-catch` in the same fashion as in the `get` function.

I will provide a pull request with this solution soon as this issue is blocker for my project.

petr-ujezdsky:
Introduced Promise API with proper error handling during setTimeout callback invocation (fixes #287), updated README accordingly

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,31 @@ new Fingerprint2().get(function(result, components){
 });
 ```
 
+Or using Promise
+
+```js
+new Fingerprint2(options).getPromise()
+  .then(function([result, components]){
+    console.log(result); //a hash, representing your device fingerprint
+    console.log(components); // an array of FP components
+  })
+  .catch(function(error){
+    console.error('Failed to compute fingerprint (probably due to some fingerprint-blocking browser extension/plugin)', error);
+  });
+```
+
+Another Promise example - it never fails and resolves with either computed fingerprint or value ('<blocked>') in case of an error
+
+```js
+new Fingerprint2(options).getOrDefaultPromise('<blocked>')
+  .then(function([result, components, error]){
+    console.log(result); // computed fingerprint or '<blocked>' in case of an error
+    console.log(components); // an array of FP components or undefined in case of an error
+    if (error) { // undefined or exception in case of an error
+      console.warn('Fingerprint probably blocked, resolved with value \'' + result + '\'', error);
+    }
+});
+
 #### You can pass an object with options (all of which are optional):
 
 ```js
--- a/fingerprint2.js
+++ b/fingerprint2.js
@@ -51,7 +51,27 @@
       }
       return target
     },
-    get: function (done) {
+    getOrDefaultPromise: function(defaultResult, defaultComponents){
+      var that = this;
+      return new Promise(function(resolve, reject) {
+        that.getPromise()
+          .then(resolve)
+          .catch(function(e) {
+            resolve([defaultResult, defaultComponents, e])
+          });
+      })
+    },
+    getPromise: function() {
+      var that = this;
+      return new Promise(function(resolve, reject) {
+        // aggregates all resolve function args into one array-like object (resolve() supports only one arg)
+        var argJoiner = function() {
+          return resolve(arguments)
+        }
+        that.get(argJoiner, reject)
+      })
+    },
+    get: function (done, reject) {
       var that = this
       var keys = {
         data: [],
@@ -102,7 +122,7 @@
         })
         var murmur = that.x64hash128(values.join('~~~'), 31)
         return done(murmur, newKeys.data)
-      })
+      }, reject)
     },
     customEntropyFunction: function (keys) {
       if (typeof this.options.customFunction === 'function') {
@@ -290,11 +310,11 @@
       }
       return keys
     },
-    fontsKey: function (keys, done) {
+    fontsKey: function (keys, done, reject) {
       if (this.options.excludeJsFonts) {
         return this.flashFontsKey(keys, done)
       }
-      return this.jsFontsKey(keys, done)
+      return this.jsFontsKey(keys, done, reject)
     },
     // flash fonts (will increase fingerprinting time 20X to ~ 130-150ms)
     flashFontsKey: function (keys, done) {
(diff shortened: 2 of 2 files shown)


fingerprint2.js
```diff
@@ -51,7 +51,27 @@
       }
       return target
     },
-    get: function (done) {
+    getOrDefaultPromise: function(defaultResult, defaultComponents){
```
jonashaag: IMO `getOrDefaultPromise` this is too special-case to be included in the library, and the code is trivial, so I'd rather not include it. Do you think it should be included?

fingerprint2.js
```diff
@@ -51,7 +51,27 @@
...
+          .catch(function(e) {
+            resolve([defaultResult, defaultComponents, e])
+          });
+      })
+    },
+    getPromise: function() {
+      var that = this;
+      return new Promise(function(resolve, reject) {
+        // aggregates all resolve function args into one array-like object (resolve() supports only one arg)
+        var argJoiner = function() {
```
jonashaag: Please use an explicit, "non-magic" function here. It's ok to "hardcode" the function signature. It helps people understand the code.

fingerprint2.js
```diff
@@ -316,181 +336,188 @@
...
-        var baseFonts = ['monospace', 'sans-serif', 'serif']
+        try {
+          that.jsFontsKeySynchronous(keys, done, that)
+        } catch(e) {
+          reject(e)
+        }
+      }, 1)
+    },
+    // kudos to http://www.lalit.org/lab/javascript-css-font-detect/
+    jsFontsKeySynchronous: function (keys, done, that) {
```
jonashaag: Can we make this function behave like any other synchronous component function, i.e. take `keys` and return `keys`? Then we could have all the promise/async related code in `jsFontsKey`.

fingerprint2.js
```diff
@@ -51,7 +51,27 @@
...
+      return new Promise(function(resolve, reject) {
+        that.getPromise()
+          .then(resolve)
+          .catch(function(e) {
+            resolve([defaultResult, defaultComponents, e])
+          });
+      })
+    },
+    getPromise: function() {
+      var that = this;
```
jonashaag: Please run standard (https://standardjs.com/) over the code. Sorry if this isn't documented yet.

fingerprint2.js
```diff
@@ -316,181 +336,188 @@
...
       var that = this
       // doing js fonts detection in a pseudo-async fashion
       return setTimeout(function () {
-        // a font will be compared against all the three default fonts.
-        // and if it doesn't match all 3 then that font is not available.
-        var baseFonts = ['monospace', 'sans-serif', 'serif']
+        try {
+          that.jsFontsKeySynchronous(keys, done, that)
+        } catch(e) {
+          reject(e)
```
jonashaag: Shouldn't there be a null check for `reject` here, or am I missing something?

README.md
```diff
@@ -60,6 +60,31 @@ new Fingerprint2().get(function(result, components){
 });
 ```
 
+Or using Promise
+
+```js
+new Fingerprint2(options).getPromise()
+  .then(function([result, components]){
```
jonashaag: Make the function signature `({result: ..., components: ...})`. This is equally simple to use in ES6 (`({ result, components }) => ...`) and it avoids magic numbers.

fingerprint2.js
petr-ujezdsky: I think it could be very common usecase so I include it. But I can move it into the README if you do not want it to be part of the library.

fingerprint2.js
petr-ujezdsky: Are you referring to the semicolon? I am disappointed that not typing the semicolon made it into some sort of standard now :( But that is for another discussion :).
Will remove it of course.

fingerprint2.js
petr-ujezdsky: Will do

fingerprint2.js
petr-ujezdsky: You are right :) By the way the "reject" parameter in get() function is not part of the public api, because errors within the get() function are not handed over to the reject().
This is on purpose to keep the current functionality and making my changes as small as possible.

fingerprint2.js
petr-ujezdsky: I do not know why you use the setTimeout function in the first place. If it was synchronous, everything would be much simpler.
I assume it is because of UI lag reason. The fonts part is probably computation-heavy and could be user unfriendly doing it synchronously.

petr-ujezdsky:
I have another big idea to make your library better. Doing it will allow to:

* make cleaner synchronous and asynchronous (promises) API 
* the error handling would be configurable (fail on first error completely, without producing any fingerprint OR only skip given component, still producing some fingerprint)
* very robust against any plugin blockers (in the sense of not failing completely)
* cleaner code

But I do not know if I would have time to implement it in my already nearly non-existing free time..

fingerprint2.js
jonashaag: Yes please!

fingerprint2.js
jonashaag: Yeah and a few other stylistic things in your patch as well. Just use the standard autoformatter :)

fingerprint2.js
jonashaag: I guess so! I'm not the original author of this project so I'm not 100% sure about the reasoning here but I'm sure there's good reasons! :)

fingerprint2.js
petr-ujezdsky: How do I run the proper linter? I have tried to run 'gulp lint', but that gave me like bazillion errors due to the not used semicolon.

jonashaag:
> I have another big idea to make your library better

Well this would be awesome and greatly appreciated but it's always a question of free time. I'm more than happy to assist, review, and discuss but I'm also not going to have much time to do coding.

jonashaag:
But yeah I guess fingerprint blockers and anti-fingerprinting built into browser is going to be an annoying topic very soon.

jonashaag:
@petr-ujezdsky are you planning on finishing this PR?

petr-ujezdsky:
I was, but I could not find free time since :( The next free time window will be around 26.3.2018. I will see if I could squeeze those minor changes into my schedule.

jonashaag:
Bump

jonashaag:
Closing in favour of #388.

PR #296, opened 2018-01-10
Add Promisify get method (final)

brendaniel:
if the get method does not have a callback function, it returns Promise

sorry for many pull request


jonashaag:
Thanks for the PR! Closing in favour of #291

PR #300, opened 2018-02-08
Change testString in font detection to enhance it's precision

ViRPo:
After playing with font detection for a while, I found out, that original testString missed one of default fonts on my machine ('Arial Hebrew' on MB Pro 2016)
It makes sense that not only size of font but number/type of characters matters a lot in this font detection strategy
Perhaps adding alphabet in both cases would be even better

What do you think?

### DIFF
--- a/fingerprint2.js
+++ b/fingerprint2.js
@@ -380,7 +380,7 @@
 
         // we use m or w because these two characters take up the maximum width.
         // And we use a LLi so that the same matching fonts can get separated
-        var testString = 'mmmmmmwwwwwwMMMMMMWWWWWWlililiLILILI111lllIII'
+        var testString = 'mMwWlLiI0123456789'
 
         // we test using 72px font size, we may use any size. I guess larger the better.
         var testSize = '72px'
--- a/fingerprint2.js
+++ b/fingerprint2.js
@@ -380,7 +380,7 @@
 
         // we use m or w because these two characters take up the maximum width.
         // And we use a LLi so that the same matching fonts can get separated
-        var testString = 'mmmmmmmmmmlli'
+        var testString = 'mmmmmmwwwwwwMMMMMMWWWWWWlililiLILILI111lllIII'
 
         // we test using 72px font size, we may use any size. I guess larger the better.
         var testSize = '72px'
--- a/fingerprint2.js
+++ b/fingerprint2.js
@@ -380,7 +380,7 @@
 
         // we use m or w because these two characters take up the maximum width.
         // And we use a LLi so that the same matching fonts can get separated
-        var testString = 'aAbBcCdDeEfFgGhHiIjJkKlLmMnNoOpPqQrRsStTuUvVwWxXyYzZ0123456789'
+        var testString = 'mmmmmmmmmmlli'
 
         // we test using 72px font size, we may use any size. I guess larger the better.
         var testSize = '72px'
--- a/fingerprint2.js
+++ b/fingerprint2.js
@@ -380,7 +380,7 @@
 
         // we use m or w because these two characters take up the maximum width.
         // And we use a LLi so that the same matching fonts can get separated
-        var testString = 'mMwWlLiI0123456789'
+        var testString = 'aAbBcCdDeEfFgGhHiIjJkKlLmMnNoOpPqQrRsStTuUvVwWxXyYzZ0123456789'
 
         // we test using 72px font size, we may use any size. I guess larger the better.
         var testSize = '72px'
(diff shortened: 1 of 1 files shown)


jonashaag:
Could you please try to explain this in other words? I'm not sure I understand what you're saying

ViRPo:
@jonashaag Sorry about that
tldr: by changing the testString from `mmmmmmmmmmlli` to `mmmmmmwwwwwwMMMMMMWWWWWWlililiLILILI111lllIII` we can improve accuracy of javascript font detection

Why is it so?

Because we check width and height of the resulting text, the original testString seems to be not various enough to cause width or height to change
As an example where my proposed testString works better is on font 'Arial Hebrew' which gets detected with my testString but doesn't with the original

However, I just realized, I was testing this on default fontList only, not on the extended, so I will check that aswell and perhaps come back with even better testString to catch all fonts

Is that clear enough? I am down for chatting about it on Skype or somewhere else if it is not clear enoguh yet :smile:

jonashaag:
I see. Your string looks pretty random to me, could you go into details why this exact sequence of characters has to be used? Like, isn't there a shorter sequence that fixes the problem?

ViRPo:
@jonashaag I updated the PR with shorter testString that gives the same result

To explain what I did and why I did it, the comment before this code says:
`// we use m or w because these two characters take up the maximum width.`
`// And we use a LLi so that the same matching fonts can get separated`
But that makes no sense if 'w' or 'm' characters have the same length and witdh in two fonts that are being compared
Instead of testing font with a long sequence of 1 char, we are much more likely to find a difference using many various characters

What I noticed is, that character '1' makes a huge difference in these tests and I guessed, that 'm', 'w', 'l' and 'i' might differ the most between two fonts

Following what I just said (that using different characters is better) it would make sense to set testString to 'aAbBcCdDeEfFgGhHiIjJkKlLmMnNoOpPqQrRsStTuUvVwWxXyYzZ0123456789' but would you be OK with such a long string?
I tested this longer testString on my system and it gives same result as 'mMwWlLiI0123456789', but that is true only for my system, the longer one should in theory work better

What do you think?

jonashaag:
@Valve opinions/insights here?

Valve:
@jonashaag I think that the intuition is right here, but before merging this, I would really want to see the before/after font detection counts on different platforms. @ViRPo  would it be too much to host a HTML on S3 bucket / github pages with before vs after so that we could have a look on various devices?

If the numbers do indeed improve and there are no false positives, then, yeah, let's merge this

ViRPo:
@Valve Sure
I was preparing quick and dirty page for you and when I tested it, to my surprise, both strings gave me the same results
I was scratching my head for a while, trying to figure out why and then found out, that Bootstrap caused it

You can check 2 versions (with and without Bootstrap) here:
https://virpo.github.io/fp2js-improve-font-detection/
https://virpo.github.io/fp2js-improve-font-detection/bootstrap

So there are 2 questions I have now:
1) Will changing the string help improve font detection results?
2) Is there anything we can do to eliminate external CSS rules from affecting font detection results? Something like CSS reset for a single DOM element perhaps?

Valve:
@ViRPo second URL not working for me. CSS reset on a single element looks like a good idea. I've briefly tested and on my linux workstation the 2 versions produce same resutls (URL #1)

ViRPo:
@Valve Ouch, try this link:
https://virpo.github.io/fp2js-improve-font-detection/bootstrap/index.html

I will do some research about that kind of CSS reset if you don't know about one already

Valve:
Thanks, working now.

You can try applying these rules in JS:
https://github.com/twbs/bootstrap/blob/v4-dev/scss/mixins/_reset-text.scss

ViRPo:
@Valve I tested the CSS reset and it worked like charm!

Then I hoped to BrowserStack and compared how new vs old testString compare, but there was no difference
However, when comparing the CSS reset with no-css reset the difference was obvious

I updated the PR with the CSS reset and reverted the testString

Valve:
thanks @ViRPo I'll have a look later today

ViRPo:
@Valve
Yo, did you have a chance to look at it by any chance? :)

Valve:
@ViRPo no.. too much work, and Jonas is the current maintainer, not me

jonashaag:
I'd like to see a quality/accuracy benchmark with as many browsers as possible. Did you do something like that on browsershots or whatever? Can I see the results?

jonashaag:
Also, how does reset only compare to reset + new string

ViRPo:
@jonashaag
I found no difference between reset + new string and reset only, therefore I removed my new string from PR

The CSS reset is important for cross-website compatibility and for the fingerprint to be immune to CSS changes on website, I see no issues that could occur cross-browser

What exactly do you want to test across browsers? I can prepare a page for it

jonashaag:
OK, that makes sense. Let's see how this goes.

PR #302, opened 2018-02-27
remove duplicate fonts from fontList

otanim:
remove font names from `fontList` in case if user defined option `userDefinedFonts` with list of values that might have duplicate values:  

#### 1st case:
```
new Fingerprint2({
  userDefinedFonts: ["Nimbus Mono", "Junicode", "Nimbus Mono"]
}).get(function(result, components) {
  console.log(result)
})
```
#### 2nd case:
```
new Fingerprint2({
  userDefinedFonts: ["Nimbus Mono", "Junicode", "Arial"]
}).get(function(result, components) {
  console.log(result)
})
```


jonashaag:
Thanks!

Issue #299, opened 2018-02-04
Code optimization


sirian:
![image](https://user-images.githubusercontent.com/897643/35776803-2f51b4aa-09b4-11e8-8174-46b9a2921cab.png)

You could easily replace this huge block of code using loops. Like this

```javascript
for (const shader of ['VERTEX', 'FRAGMENT']) {
    for (const numType of ['FLOAT', 'INT']) {
        for (const numSize of ['HIGH', 'MEDIUM', 'LOW']) {
            for (const key of ['precision', 'rangeMin', 'rangeMax']) {
                const format = gl.getShaderPrecisionFormat(gl[shader + '_SHADER'], numSize + '_' + numType);
                result.push(`webgl ${shader} shader ${numType} ${numSize} ${key}:` + format[key]);
            }
        }
    }
}

Valve:
thanks @sirian 
@jonashaag  what do you think?

jonashaag:
Yeah we could merge that but I think the value it provides is rather low. And I'd like to see a PR that doesn't use ES6 features since we're not using ES6 in the library so far. So +0 from me if we have a PR.

Valve:
thanks, I agree

sirian:
@jonashaag I'm not talking about performance. I mean this approach significantly reduces file size (for example for this block - 5kb vs 400bytes)
And also this block may be optimized 
![image](https://user-images.githubusercontent.com/897643/36248689-66ea212a-1249-11e8-9adb-f0645f908259.png)

Unfortunately I haven't enough free time to make PR(

ProcrastinatorCp:
@Valve , @jonashaag 
Hi, I would like to give it a try.

jonashaag:
Cool! Im happy to assist.

namanjain1812:
@jonashaag @Valve 
hi,is this issue solved?Would like to contribute if open

ProcrastinatorCp:
@namanjain1812 , I assumed you were not working on it, therefore I created this PR. 

@jonashaag, 
Could you please review https://github.com/Valve/fingerprintjs2/pull/317 and comment.

PR #317, opened 2018-04-29
#299 Code Optimization - replace hard-coded `webgl format` with loop

ProcrastinatorCp:
## Description

- Code Optimization. (Using loop to generate `webgl shader precision` )
- Added `error` log on `uglify` 

## Related Issue
https://github.com/Valve/fingerprintjs2/issues/299

## Motivation and Context

Reducing the file size of the block changed. 
- File size before : **36.2 kB** 
- File size after : **32.1 kB**

## Types of changes

- [x] Code Optimization
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** document.
- [x] All new and existing tests passed.
- [ ] My change requires a change to the documentation.


package.json
```diff
@@ -37,5 +37,8 @@
       "specs/lib",
       "specs/phantomjs-testrunner.js"
     ]
+  },
+  "dependencies": {
+    "uglify-es": "^3.3.9"
```
ProcrastinatorCp: added dependency for es6

gulpfile.js
```diff
@@ -24,6 +26,8 @@ gulp.task('minify', function () {
         output: {
           ascii_only: true
         }
+      }).on('error', function(e) {
```
ProcrastinatorCp: While debugging, the errors where not shown with file name and line number. Added error for easier debugging. Let me know if not needed.

gulpfile.js
```diff
@@ -1,7 +1,9 @@
 const gulp = require('gulp')
 const standard = require('gulp-standard')
 const rename = require('gulp-rename')
-const uglify = require('gulp-uglify')
+const uglifyes = require('uglify-es')
```
ProcrastinatorCp: using `uglify-es` for ES6.

fingerprint2.js
```diff
@@ -899,6 +871,31 @@
         return null
       }
     },
+    getWebGlShaderPrecisionFormat: function () {
+      try {
+        var gl = this.getWebglCanvas()
+        if (!gl) { return null }
+        var resultArray_ = []
+        for (const numType of ['FLOAT', 'INT']) {
```
jonashaag: Don't use `for ... of` loops, it's not supported by all browsers. My suggestion is to keep the unfolded, hard-coded list of combinations, i.e.

```
[
['FLOAT', 'VERTEX', 'HIGH', 'precision'],
['FLOAT', 'VERTEX', 'HIGH', 'rangeMin'],
...
['FRAGMENT', 'LOW', 'rangeMax']
]
```

Or use `.forEach`, which is polyfilled for browsers that don't support it.

gulpfile.js
jonashaag: Remove this; we're not using ES6 yet.

package.json
jonashaag: see above

jonashaag:
Please don't recompile .min.js

ProcrastinatorCp:
Hi @jonashaag 
Thanks for reviewing the PR. I have fixed as per the feedback. 
> Please don't recompile .min.js
I couldn't understand the above. Would you please explain.

fingerprint2.js
```diff
@@ -899,6 +870,34 @@
         return null
       }
     },
+    getWebGlShaderPrecisionFormat: function () {
```
jonashaag: inline this function into the place where it's called, and get rid of the array copying. Directly push to the result array.

specs/specs.js
```diff
@@ -261,6 +261,44 @@ describe('Fingerprint2', function () {
       })
     })
 
+    describe('webgl shader precision format', function () {
+      it('checks webgl shader precision format loop', function (done) {
+
```
ProcrastinatorCp: Checking if all the corresponding entries are present. 
Couldn't think of a better approach for this spec. Please suggest if any.

ProcrastinatorCp:
> If you're interested you can also try to make the series ... 

On a first scan, the block doesn't seem to be too similar to reduce. Will give it a try in near future.

specs/specs.js
jonashaag: The problem with this kind of test is that it duplicates the code logic 1:1. Tests shouldn't try to be smart, as every little bit of smartness increases the chance of a bug in the tests.

My suggestion: Copy and paste the list of entries from the old fingerprint.js file into the tests, and compare them. It's find to have a hardcoded list of entries in the tests.

ProcrastinatorCp: Cool. Done.

jonashaag:
Merged with some minor changes, thanks!

sirian:
@ProcrastinatorCp 
You could also refactor this part if you wish)
image

for example

```
result.push('webgl max cube map texture size:' + gl.getParameter(gl.MAX_CUBE_MAP_TEXTURE_SIZE))
      result.push('webgl max fragment uniform vectors:' + gl.getParameter(gl.MAX_FRAGMENT_UNIFORM_VECTORS))
      result.push('webgl max render buffer size:' + gl.getParameter(gl.MAX_RENDERBUFFER_SIZE))
```


```
var wql = [
['max cube map texture size:', gl.MAX_CUBE_MAP_TEXTURE_SIZE],
['max fragment uniform vectors:',gl.MAX_FRAGMENT_UNIFORM_VECTORS],
['max render buffer size:', gl.MAX_RENDERBUFFER_SIZE]
...
]

this_.each(wgl, function (pair) {
    result.push('webql ' + pair[0], gl.getParameter(pair[1]))
});

jonashaag:
You could even infer the human readable name from the constant name

PR #327, opened 2018-06-01
getAdBlock returns null instead of false

plevy:
I do not know the exact condition causing this but result is being set to null and returning null rather than false.  To be consistent and prevent the index.html example from sending a null to .toString, check if null, and set to false if necessary.


jonashaag:
Thanks for the patch, but that's impossible, the error must be somewhere in your code.

PR #336, opened 2018-07-06
Fixed haslieos when user use iPad or iPhone

nian0114:


jonashaag:
Thanks for the patch. Can you add a test for this bug?

jonashaag:
Closing due to inactivity.

PR #346, opened 2018-07-19
Resolved #345

### LINKED ISSUE #345: TypeError: {(intermediate value)} is not a function
### Problem description

When I run tests, I got an error as follows:  
TypeError: {(intermediate value)} is not a function

### Fingerprintjs2 version and devices used

- I can reproduce the bug in fingerprintjs versions: newest
- I can reproduce the bug on these devices: Chrome

wkdtjsgur100:
## Related Issue

#345

### DIFF
--- a/specs/specs.js
+++ b/specs/specs.js
@@ -10,6 +10,8 @@ function getComponent(components, key) {
   throw new Error('Component not found: ' + key)
 }
 
+onPhantomJs = window._phantom != null
+
 describe('Fingerprint2', function () {
   describe('new', function () {
     it('creates a new instance of FP2', function () {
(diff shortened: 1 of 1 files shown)


jonashaag:
Thanks for always filling out the issue template, I've added a note about not having to use it if you're submitting a PR.

Thanks for fixing this. Did you happen to know why this broke in the first case? Should be valid JS shouldn't it?

wkdtjsgur100:
https://stackoverflow.com/questions/42036349/uncaught-typeerror-intermediate-value-is-not-a-function

I think you didn't use the semicolon. 
Here is a Note:
https://stackoverflow.com/questions/42036349/uncaught-typeerror-intermediate-value-is-not-a-function

jonashaag:
Oh, standardjs should have complained about this, but looks like it's missing from the CI pipeline!

PR #347, opened 2018-07-20
 Add maxLength, maxLengthSalt options for performance improvement

wkdtjsgur100:
- Add maxLength, maxLengthSalt option descriptions in README.md
- Add getSubstrByMaxLength, getSalt function
- Add test code


jonashaag:
Thanks for bringing this up.

What's the rationale behind the salting?

wkdtjsgur100:
Salting is used to determine the starting point when trimming a string. (maxLength option means trimming from starting point to starting point + maxLength)

For example, if you change the UserAgent using salt, you can also set the cutting off of the string. This increases the uniqueness as compared to trimming the string from 0 to maxLength.

Issue #348, opened 2018-07-20
 Add maxLength, maxLengthSalt options for performance improvement


wkdtjsgur100:
# Problem description

In some cases, if the length of the canvas and webgl fingerprint are too large, it can cause performance degradation. (This is because it causes problems that take too long to generate a hash.)  To prevent this, `maxLength` option can limit the length of the canvas and webgl fingerprint.

Pull request is here.
[#347](https://github.com/Valve/fingerprintjs2/pull/347)

Please read README.md I modified in detail.

jonashaag:
Closing due to inactivity.

PR #351, opened 2018-07-23
Add maxLength, maxLengthSalt options for performance improvement 

wkdtjsgur100:
I'm sorry, git was twisted, that's why I send you PR again.

# Previous pull request

#347 
 
# Changing log
- Add maxLength, maxLengthSalt option descriptions in README.md
- Add getSubstrByMaxLength, getSalt function
- Add test code

# related issue

#348


jonashaag:
Great idea. I'm not happy with basing the salting length on user agent though. That way in the (very unlikely) case that you have two different browsers with identical, overly long canvas/GL fingerprints, fingerprints will be different, which makes canvas/GL unnecessarily browser-dependent.

Not sure what an alternative is, give me a few days to think about this or make a suggestion

wkdtjsgur100:
My another suggestion is to apply string compression algorithm that is faster than hash algorithm. However I don't know efficient compression algorithm yet. If I find out it I'll let you know.  Thank you.

jonashaag:
That would be best. How long are the "long" strings you're seeing compared to the "average" strings?

We could also decide the cut position from hashing a few bytes of the canvas/GL fingerprint itself, which makes the cut position not depend on the UA. Maybe not the first few bytes, but maybe bytes 100-110 or whatever.

jonashaag:
Closing due to inactivity.

PR #354, opened 2018-07-25
Rename excludeAudioFP -> excludeAudio

RobertJGabriel:
## Changing log
- Rename excludeAudioFP -> excludeAudio

## related issue

https://github.com/Valve/fingerprintjs2/issues/353


RobertJGabriel:
@jonashaag

jonashaag:
Thanks, I've applied it to the source only

PR #359, opened 2018-08-03
Add measurePerformance option

wkdtjsgur100:
As you can see the test code, `measurePerformance` option adds time key on each component.

I think you can get a detail report related to performance if you add this feature.

Thank you


jonashaag:
Great idea.

I'd rather have the benchmarking code separate from the implementation.

How about implementing this with a small script that excludes all options except a single one, measures the time, and repeats that process for all options? Theoretically we'd have to account for the time spent in hashing that is not proportional to the size of the hashing function input (i.e., fixed upfront overhead for calling the hashing function), but that's probably negligibly low.

wkdtjsgur100:
Good. I'll try it.

wkdtjsgur100:
@jonashaag By the way, What's the reason why you want to separate code from the implementation?

wkdtjsgur100:
And It could be dropped the request if I generate a lot of fingerprint object on some case

jonashaag:
Which request?

The same reason you want to separate test code from the implementation. Separation of concerns. You should never intertwine implementation and tests/benchmarks/whatever

wkdtjsgur100:
Actually, I'm testing on the actual operating environment with a fingerprint, and I want to measure the performance based on it. But in some cases, I found out many requests could be dropped because some users are leaving while the fingerprint is loaded. (Fingerprint is too slow in some situation. For example, IE and windows 7 environment.)

Anyway, I totally understood why you want to separate benchmark codes from implementation. I'm going to try to implement what you told me.

jonashaag:
Closing due to inactivity.

Issue #356, opened 2018-07-30
Suggestion: Add timezone from the Internationalization API if available


SleepProgger:
The [Internationalization API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl) provides, besides some other maybe interesting data, the time zone the user is in.
This data contains more information as a simple `new Date().getTimezoneOffset()` does.

Example from my system:
`Intl.DateTimeFormat().resolvedOptions()`
->
`Object { locale: "en-US", calendar: "gregory", numberingSystem: "latn", timeZone: "Europe/Berlin", month: "numeric", day: "numeric", year: "numeric" }`

jonashaag:
PR welcome

SleepProgger:
Awesome, i'll send an PR sometime the next week.
I assume only the timeZone and disabled by default to not suddenly produce different hashes for the same user ?

SleepProgger:
Added in #364

PR #364, opened 2018-08-08
Added timezone from the Internationalization API

SleepProgger:
This PR adds the timezone from `Intl.DateTimeFormat().resolvedOptions().timeZone`.
For most systems/browser this will return the timezone in the  IANA time zone database format, fe: "Europe/Berlin" or "America/Los_Angeles".

Disabled by the "excludeTimezone" option.

More info:
[Intl.DateTimeFormat](https://www.ecma-international.org/ecma-402/1.0/#sec-12)
Ref. Issue: #356

### DIFF
--- a/fingerprint2.js
+++ b/fingerprint2.js
@@ -81,6 +81,7 @@
       keys = this.screenResolutionKey(keys)
       keys = this.availableScreenResolutionKey(keys)
       keys = this.timezoneOffsetKey(keys)
+      keys = this.timezoneKey(keys)
       keys = this.sessionStorageKey(keys)
       keys = this.localStorageKey(keys)
       keys = this.indexedDbKey(keys)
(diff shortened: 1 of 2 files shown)


jonashaag:
Thanks!

Can you add a positive test case, maybe simply check if the time zone name follows IANA format (does it contain a /, or whatever).

SleepProgger:
Not as far as i know. Something like UTC or GMT+2 are also valid values.

jonashaag:
Ok then maybe simply check for non-empty.

SleepProgger:
As in: check if the entry is in the keys array ?

Or if the value is empty ?
Because FF 49 for example returns undefined as the default timeZone. Not sure if some browser version would also return ''. In my opinion as long as the browser implements that API we should use that value, whatever it is. Even if it is undefined that is some more information about the browser after all. (Although stringifying the result would maybe a good idea ?)


Side note: maybe there should be some more generalized test for exclude{CHECK} options in the specs ? Assuming the naming convention is consistent (option name to exclude checks always = 'exclude' + capitalize(components[x].key)) that should be easy to add.

jonashaag:
> Because FF 49 for example returns undefined as the default timeZone. Not sure if some browser version would also return ''. In my opinion as long as the browser implements that API we should use that value, whatever it is. Even if it is undefined that is some more information about the browser after all. (Although stringifying the result would maybe a good idea ?)

I see, then let's not add a test. I wanted a test to make sure the feature works, but if it's impossible to test if it's work then it's impossible to add a test. :)

> Side note: maybe there should be some more generalized test for exclude{CHECK} options in the specs ? Assuming the naming convention is consistent (option name to exclude checks always = 'exclude' + capitalize(components[x].key)) that should be easy to add.

Yeah. We should refactor a lot of things in the library and the test suite. If you're interested in refactoring some stuff in the library we can chat about that!

jonashaag:
Can you add the new option to https://github.com/Valve/fingerprintjs2/wiki/List-of-options

SleepProgger:
Sure thing.  Done.

PR #378, opened 2018-09-03
Check for existance of `window` so we don't break server side render

alexhayes:
I was having issues `import`ing fingerprint2 when using using server side rendering.

I've created an example repo at https://github.com/alexhayes/fingerprint2-broken-ssr which demonstrates this issue.

### DIFF
--- a/fingerprint2.js
+++ b/fingerprint2.js
@@ -18,7 +18,7 @@
 
 (function (name, context, definition) {
   'use strict'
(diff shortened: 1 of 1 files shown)


jonashaag:
If you could add an automated test to check we don't re-introduce this bug in the future, that would be awesome.

alexhayes:
Thanks for merging so quickly @jonashaag 

Regarding a test I did have a look at this but it looked like the test runner was pretty tied to phantomjs and thus wasn't too sure how to make that happen. 

Any thoughts how to do this without a major refactor to the test runner?

jonashaag:
We can always add an additional test script to the suite. Maybe not as elegant but better than not having a test.

PR #383, opened 2018-09-19
Code quality Semver Backwards compatible changes

GrosSacASac:
```
use valid html
rename oncompleteTimeout
refactor: remove unnecessary keys reassignement
Give example of setTimeout  …
Refactor index.html, avoid innerHTML, fingerprint on load, avoid escapeHTML hack
1.8.0 -> 1.8.2
```


GrosSacASac:
should we update gulp-util ? In a separate PR ?

> npmWARNdeprecated gulp-util@3.0.8: gulp-util is deprecated - replace it, following the guidelines at https://medium.com/gulpjs/gulp-util-ca3b1f9f9ac5

I see that travis is configured with:
```
node_js:
  - 5
  - 6
  - 7
  - 8
  - 9
  - 10
```
We should consider droping 5, 7 , 9 as NodeJs organization dropped them

index.html
```diff
@@ -66,48 +65,50 @@
...
 
-
   <h1>Fingerprintjs2</h1>
 
-  <p>Your browser fingerprint: <strong id="fp"></strong></p>
-  <p><code id="time"/></p>
-  <p><span id="details"/></p>
   <button type="button" id="btn">Get my fingerprint</button>
-
-  <a href="https://github.com/Valve/fingerprintjs2"><img style="position: absolute; top: 0; right: 0; border: 0;" src="https://camo.githubusercontent.com/365986a132ccd6a44c23a9169022c0b5c890c387/68747470733a2f2f73332e616d617a6f6e6177732e636f6d2f6769746875622f726962626f6e732f666f726b6d655f72696768745f7265645f6161303030302e706e67" alt="Fork me on GitHub" data-canonical-src="https://s3.amazonaws.com/github/ribbons/forkme_right_red_aa0000.png"></a>
```
jonashaag: Why remove this?

README.md
```diff
@@ -35,13 +35,16 @@ This project uses `semver`.
 ## Usage
 
 ```js
-new Fingerprint2().get(function(result, components) {
-  console.log(result) // a hash, representing your device fingerprint
-  console.log(components) // an array of FP components
-})
+setTimeout(function () {
```
jonashaag: Would requestIdleCallback be as good as setTimeout here? Should we recommend it instead?

jonashaag:
> should we update gulp-util ? In a separate PR ?
> 
> > npmWARNdeprecated gulp-util@3.0.8: gulp-util is deprecated - replace it, following the guidelines at [medium.com/gulpjs/gulp-util-ca3b1f9f9ac5](https://medium.com/gulpjs/gulp-util-ca3b1f9f9ac5)

yes, this PR is ok

> 
> I see that travis is configured with:
> 
> ```
> node_js:
>   - 5
>   - 6
>   - 7
>   - 8
>   - 9
>   - 10
> ```
> We should consider droping 5, 7 , 9 as NodeJs organization dropped them

Agreed.

index.html
GrosSacASac: Accidental sorry, re added it

README.md
GrosSacASac: requestIdleCallback has its uses, but for now would not recommend it over setTimeout, as it is still a draft

GrosSacASac:
Ok should be good now :)

README.md
jonashaag: Why use it in the demo HTML then?

GrosSacASac: requestIdleCallback executes as soon as the browser is idle without having to guess a good timing with setTimeout. It also executes after all the other setTimeouts without having to know their timings because it has lower priority.

It basically is a "execute as soon as possible, after everything else has loaded and executed".

But the downside is that it is less predictable. For example if you make a request that should contain the fingerprint in a setTimeout of 5 seconds the IdleCallback may have executed, however if fingerprint was in a setTimeout of 500ms you have correct order guarantee.

To get best of both worlds, one could use an [idle-until-urgent helper](https://philipwalton.com/articles/idle-until-urgent/)

GrosSacASac: Of course it depends how the application is coded.

The main purpose of our html file is to display the fingerprint and that's it.

Of course setTimeout is still used as a fallback.

GrosSacASac: I just checked https://www.w3.org/TR/requestidlecallback/ and it not a draft anymore but a Proposed Recommendation.

So we could recommend it in the readme also, what do you think ?
 * requestIdleCallback **optimal** **may require application coordination** **only firefox and chrome**
 * setTimeout **stable** **not optimal**

jonashaag: Yes, and add a comment to the demo HTML that people should go read the comment in the README.

GrosSacASac: done

PR #414, opened 2018-12-18
Fixed problem with loading library on pages where exists require.js

pbylina:
but outside of require.js context https://requirejs.org/docs/errors.html#mismatch


jonashaag:
Thank you

PR #433, opened 2019-02-23
Add navigator.webdriver to components

bes-internal:
Good and stable signal.

Ref:
https://www.chromestatus.com/feature/6216034532982784
https://developer.mozilla.org/en-US/docs/Web/API/Navigator/webdriver


jonashaag:
Thank you!

PR #436, opened 2019-03-05
add guard against string value being treated as array

serkan-paystack:
This PR addresses https://github.com/Valve/fingerprintjs2/issues/432 where canvas might possibly be throwing and returning an error string instead of an expected error.

The hotfix simply guards against the value not being an array and instead continues with the program flow.

This does not necessarily fix the underlying root cause, merely ensures it does not break full execution.

### DIFF
--- a/fingerprint2.js
+++ b/fingerprint2.js
@@ -1,5 +1,5 @@
 /*
-* Fingerprintjs2 2.0.6 - Modern & flexible browser fingerprint library v2
+* Fingerprintjs2 2.0.7 - Modern & flexible browser fingerprint library v2
 * https://github.com/Valve/fingerprintjs2
 * Copyright (c) 2015 Valentin Vasilyev (valentin.vasilyev@outlook.com)
 * Licensed under the MIT (http://www.opensource.org/licenses/mit-license.php) license.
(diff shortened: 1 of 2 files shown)


serkanatpaystack:
Hey @jonashaag pinging you here as you're already familiar with this use case.

nebi-frame:
why are we not merging this ? PR seems promising and seems passed all the checks?

serkanatpaystack:
@nebi-frame fwiw we migrated away from the 1.8 API and embraced the new api.

Valve:
@jonashaag please review if you have time, alternatively I'll review later this week

fingerprint2.js
```diff
@@ -1388,14 +1388,9 @@
...
-          newComponents.push({key: component.key, value: component.value.join('~')})
+          if (component.value && typeof component.value.join === 'function') newComponents.push({key: component.key, value: component.value.join('~')})
         } else if (['sessionStorage', 'localStorage', 'indexedDb', 'addBehavior', 'openDatabase'].indexOf(component.key) !== -1) {
-          if (component.value) {
-            newComponents.push({key: component.key, value: 1})
-          } else {
-            // skip
-            continue
-          }
+          if (component.value) newComponents.push({key: component.key, value: 1})
```
jonashaag: @Valve I'd undo this change.

fingerprint2.js
```diff
@@ -1388,14 +1388,9 @@
               return [p[0], p[1], mimeTypes].join('::')
             })})
         } else if (['canvas', 'webgl'].indexOf(component.key) !== -1) {
-          newComponents.push({key: component.key, value: component.value.join('~')})
+          if (component.value && typeof component.value.join === 'function') newComponents.push({key: component.key, value: component.value.join('~')})
```
jonashaag: @Valve I think this should have an else case (just add `value: component.value`)

Valve: sounds good, thanks

Valve:
this has been implemented in https://github.com/Valve/fingerprintjs2/commit/d9d4b0dd266d57d53a06e05aa6356a1513fcf7da
closing.

PR #499, opened 2019-12-03
Various changes dealing with false positives on hasLiedOs() and hasLiedBrowser()

mads1980:
This PR deals with the following issues:

hasLiedOs meta issue
https://github.com/Valve/fingerprintjs2/issues/497

hasLiedOs is unreliable and can be easily improved
https://github.com/Valve/fingerprintjs2/issues/493

hasLiedBrowser Opera-check is unreliable
https://github.com/Valve/fingerprintjs2/issues/494

hasLiedBrowser is misdetecting pre-4.4 Android stock browser (and WebView) as Safari
https://github.com/Valve/fingerprintjs2/issues/495

hasLiedBrowser is unreliable for Windows Phone Edge and Opera Mini
https://github.com/Valve/fingerprintjs2/issues/496

has lied OS & has lied resolution accuracy
https://github.com/Valve/fingerprintjs2/issues/396

Discussion of the changes:

1) Substring matches for "win" that attempt to detect Windows have been replaced with more specific substrings, in order to avoid detecting anything else containing "win" as "Windows". This is very prevalent in mobile devices that contain vendor name, device model, build, etc.

2) Same as above for "mac" -> replaced with "macintosh" and "mac_powerpc"

3) X11 is added as substring match for Linux.

4) iPod, CriOS (Chrome for iOS) and FxiOS (Firefox for iOS) are added as matches for iOS

5) Windows and CrOS (Chrome OS) are added as valid touch device OS options. Avoids false positives for touch-enabled laptops.

6) Avoids false positives for Windows Phone with platform = "ARM"

7) Avoids false positives for Opera Mini with platform containing "Pike"

8) platformIsOther check takes iPod into account.

9) hasLiedBrowser() avoids checks for Microsoft Edge and Opera Mini, since both use different rendering and JavaScript engines, depending on browser version and execution mode.

10) hasLiedBrowser() uses substrings with slashes to improve matching. Especially important is the change from "opr" to " opr/" to avoid misdetecting Opera when the User-Agent contains "opr"

11) hasLiedBrowser() now attempts to differentiate AOSP (old Android stock browser) which was previously detected as "Safari" even through it is much closer to Chrome (it is based on Chromium source code). The eval.toString().length check for AOSP should be the same as Chrome, however the navigator.productSub check was left out for AOSP, since we cannot be sure what these old devices returned for navigator.productSub without an actual device test (anyone care to test?)

### DIFF
--- a/fingerprint2.js
+++ b/fingerprint2.js
@@ -1121,25 +1121,25 @@
     // We extract the OS from the user agent (respect the order of the if else if statement)
     if (userAgent.indexOf('windows phone') >= 0) {
       os = 'Windows Phone'
-    } else if (userAgent.indexOf('win') >= 0) {
+    } else if (userAgent.indexOf('windows') >= 0 || userAgent.indexOf('win16') >= 0 || userAgent.indexOf('win32') >= 0 || userAgent.indexOf('win64') >= 0 || userAgent.indexOf('win95') >= 0 || userAgent.indexOf('win98') >= 0 || userAgent.indexOf('winnt') >= 0 || userAgent.indexOf('wow64') >= 0) {
       os = 'Windows'
     } else if (userAgent.indexOf('android') >= 0) {
       os = 'Android'
-    } else if (userAgent.indexOf('linux') >= 0 || userAgent.indexOf('cros') >= 0) {
+    } else if (userAgent.indexOf('linux') >= 0 || userAgent.indexOf('cros') >= 0 || userAgent.indexOf('x11') >= 0) {
       os = 'Linux'
-    } else if (userAgent.indexOf('iphone') >= 0 || userAgent.indexOf('ipad') >= 0) {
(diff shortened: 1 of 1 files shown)


jonashaag:
@valve please review as well

jonashaag:
Looks good to me. To be honest however I'm not really familiar with the has-lied checks and also can't test all of these browsers.

Btw @mads1980 how did you test these things?

mads1980:
We are using the hasLied datapoints in a mobile antifraud solution which processes millions of HTTP requests per day. We've been tracking the hasLied values for over a year (so we have quite a big dataset from which to infer criteria), and have been actively using them for blocking-decisions for the past few weeks, using our patched branch.

Our patch mostly deals with mobile-specific issues in the current code because of our use case.

Valve:
@mads1980 thanks for the PR!
I'm a bit busy now, but will definitely review it next week!

Valve:
thanks you @mads1980!

Issue #486, opened 2019-10-10
Add automatic testing for various browsers


jonashaag:
Let’s start very simple: should run script on a static page and make sure we get hashed fingerprint and all of the components.

PR #503, opened 2019-12-10
[browser-tests] added more browsers for automated testing:

Valve:
* Firefox headless
* Firefox incognito mode
* Chrome incognito mode
* Safari

Because only Chrome/Firefox headless are supported in the CI env (on linux),
it's only possible to run yarn test:all on a Mac.
On linux, it's possible to run yarn test:chrome* and yarn test:firefox*


Valve:
fixes #486 (well, almost)

PR #523, opened 2020-04-19
add synchronous mode using {synchronous: true} in option

SeyyedKhandon:


Valve:
Hi @SeyyedKhandon 

Thanks for your pull request.

Unfortunately we cannot merge it, because in general most JavaScript tools are asynchronous and adding a synchronous version is against this async philosophy. 
Async versions are better because they will not block the event loop and in general should be preferred.

Please feel free to offer your ideas or discuss further.

Closing for now.

SeyyedKhandon:
Hi @Valve 

You're welcome, I had this need like some other developers, in some situation to use this great library as synchronous mode, this will only disable asynchronous if a user adds the `synchronous: true` to the options,  So I think it's not a bad idea to give the developer a choice to use it or not.

Issue #524, opened 2020-04-20
Fingerprint2 is undefined when using require JS


AmirSasson:
### Scenario

- use the CDN fingerprint js,  in addition add the script tag of require js
like 
```
<script src='https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.js'></script>
```
Fingerprint2 is undefined.
apparently, the condition 
```
 if (typeof window !== 'undefined' && typeof define === 'function' && define.amd) {
        define(definition)
    } 
```
is true,  and `define()` is running, after it runs i have no access to Fingerprint2  object.

can you please tell how can i get access to the **FIngerprint2**  if using `define()`
in addition, why do even use it and not just directly creating the object like in 
```
  context[name] = definition()
```
?

pisaq:
Still not working on my side :( I'm using it in Magento Store.

AmirSasson:
> Still not working on my side :( I'm using it in Magento Store.

The PR hasnt been accepted yet. Maybe u can ping the repo owner to merge it.

pisaq:
> > Still not working on my side :( I'm using it in Magento Store.
> 
> The PR hasnt been accepted yet. Maybe u can ping the repo owner to merge it.

I've took the code from PR, so fix here did not worked for me.

Valve:
guys, I'm sorry for such a long delay, I'm so much swamped by work.
Will definitely review all pending PRs, including yours @AmirSasson

AmirSasson:
So obviously you have a different problem.
What error/behavior do u have?

pisaq:
> So obviously you have a different problem. What error/behavior do u have?
> […](#)

`ReferenceError: Can't find variable: Fingerprint2`

pisaq:
@Valve, we believe in you ! ;-)

pisaq:
@AmirSasson I've fixed it by doing this 
https://github.com/Valve/fingerprintjs2/issues/320#issuecomment-633024303

Valve:
v3 will be implemented in TypeScript with good module support.

Finesse:
@AmirSasson @pisaq It looks strange. Could you please share your HTML code or the parts of the code that load and run scripts?

- If you have an HTML like this:
    ```html
    <script src="https://cdn.jsdelivr.net/npm/requirejs@2.3.6/require.min.js"></script>
[...]

Issue #469, opened 2019-06-06
Fingerprint2 is undefined inside an Electron app


F0RIS:
### Scenario

- Browser: Electron/4.1.0
- Version: 2.1.0

Initialization checks works incorrectly inside Electron. This is because you actually have "module" thing inside it and these checks do 
```module.exports = definition() ```
 instead of desired 
```context[name] = definition() ```

Here is my naive fix which is OK in my case, but not OK for the lib)
![image](https://user-images.githubusercontent.com/10178982/59022319-36d4ee00-8856-11e9-82e8-5f7c37a8c596.png)

To test it you can create simple Electron app with the code below and open any web site which uses Fingerprint2
```
var electron = require("electron");
var { app, BrowserWindow } = electron;

var mainWindow;

app.on("ready", function () {
    mainWindow = new BrowserWindow({
        width: 1280 + 400,
        height: 720 + 400,
        darkTheme: true,
    });

    mainWindow.webContents.session.clearCache(function () {
        mainWindow.loadURL("http://localhost:4200");
    });
});
``` 

I hope you can do a nice fix so this lib will work in Electron out of the box. Thanks.

jonashaag:
I'm fine with that workaround, but maybe there's a proper way to solve this? How do other libraries make sure module initialisation is done correctly? (Without using any transpilers/preprocessors etc; at some point we should simply switch to modern JS)

cekvenich:
In my code I wait for either 'deviceready' (electorn) or DOMContentLoaded (regular). Not sure if it will give you what you want.
But using modern .js is bad since ie11 has huge market share. Using typescript is cool, since it will do both es5 or es2016.

Valve:
this potentially can be solved by this initiative: https://github.com/Valve/fingerprintjs2/issues/483
Once we get a normal module support, we should be able to solve Electron issues.

cekvenich:
The problem w/ module support is that it can't be used of the CDN w/o packing.

Valve:
@cekvenich yeah, once we have modules, we can generate a UMD module which will work on CDN. 
Essentially there will be two types of modules in the distribution: UMD (for CDN) and CommonJS (for Webpack, Browserify etc).

AmirSasson:
I have kinda related scenario where the fingerprint is loaded dynamically from CDN on a browser on a site that includes a `requirejs`.
in this case it seems like i cant access the FIngerprint2 objects that usually available on the window scope.
[...]

PR #525, opened 2020-04-20
Support Initialization options 

AmirSasson:
* disable using available js module
* override default scope
* override default name

this targets a case when having a `require` in the site, and the **fingerprint** is not available globally on the window, these changes are NON breaking changes and supports the current behavior.
#524  #469

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ fingerprintjs2.iml
 *.log
 package-lock.json
 dist/
+test.html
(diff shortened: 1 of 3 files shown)


Valve:
Hello @AmirSasson 

Thanks for your pull request. 

We are planning to release a new version (v3), in TypeScript that will have wide module support.

Please stay tuned.

Closing this PR, please feel free to offer any ideas or discuss further.

PR #532, opened 2020-07-09
Improve hardwareConcurrency detection in Safari

kalexey89:
Note from [Can I use](https://caniuse.com/#search=hardwareConcurrency) for `navigator.hardwareConcurrency`:
> WebKit browsers clamp the maximum value returned to 2 on iOS devices and 8 on all others. Disabled in Safari behind the ENABLE_NAVIGATOR_HWCONCURRENCY build option.

.

### DIFF
--- a/fingerprint2.js
+++ b/fingerprint2.js
@@ -849,6 +849,15 @@
     if (navigator.hardwareConcurrency) {
       return navigator.hardwareConcurrency
     }
+
+    // Disabled in Safari behind the ENABLE_NAVIGATOR_HWCONCURRENCY build option.
+    var isSafari = navigator.userAgent && /^((?!Chrome|Android).)*Safari/i.test(navigator.userAgent);
+    if (isSafari) {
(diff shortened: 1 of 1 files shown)


Valve:
This project is not for hardware concurrency detection, it's for browser fingerprinting. This change doesn't improve browser fingerprinting accuracy, so closing it. Please feel free to reopen with additional comments.

kalexey89:
@Valve Can the fact that we can get different values ​ ​ depending on the platform help us increase the percentage of user recognition? Safari in 99% of cases does not return anything when reading this parameter, we emulate it - we can get at least different value for different platforms.

jonashaag:
I don’t see how that makes sense. 

If the user hasn’t enabled HW concurrency manually, Safari will not report anything and the component will be a mere “is it Safari?” check that you also have using UA etc. With your patch, the component will be equivalent to “is it iOS?”/“is it Mac?” which is also a subset of UA. So, you lose no information in either implementation (with or without your patch).

If the user HAS enabled HW concurrency, then in your case you lose very valuable information because having enabled that flag will be extremely rare.

PR #540, opened 2020-08-04
pref: Excludes device memory by default.

DemonLms:
Modify defaultOptions to exclude device memory by default.
Only a few browsers (such as Chrome, Edge) support this property,
but the results are different for HTTPS and HTTP website.
The browser will only return this value for trusted access like
HTTPS, localhost, or 127.0.0.1.

Compatibility: Different browser fingerprints will be generated by
default.

### DIFF
--- a/fingerprint2.js
+++ b/fingerprint2.js
@@ -269,6 +269,8 @@
     excludes: {
       // Unreliable on Windows, see https://github.com/Valve/fingerprintjs2/issues/375
       'enumerateDevices': true,
+      // Only a few browsers (such as Chrome, Edge) support this property, but the results are different for HTTPS and HTTP website.
+      'deviceMemory':true,
       // devicePixelRatio depends on browser zoom, and it's impossible to detect browser zoom
       'pixelRatio': true,
       // DNT depends on incognito mode for some browsers (Chrome) and it's impossible to detect incognito mode


Valve:
We're actually thinking about removing support for HTTP. 
In general we should not use HTTP websites and removing a source of entropy for http only websites looks moot to me.
Happy to discuss

DemonLms:
Thank you for your reply.

I don't mean remove, but exclude by default, in exchange for better compatibility, stability, and ease of use, and those who need this property can turn it back.

I'm sure HTTPS is more secure, and is the future. Our site is all HTTPS, but I can't ensure that all partners are HTTPS. Browser fingerprint is important means of cross-site tracking, but we don't know target site is HTTPS or HTTP, the fact is that we lost, because when across the protocol will have different fingerprint.

As an open source project we should not be deciding for our users whether to use HTTPS or HTTP. Millions of HTTP sites are still running, millions of HTTP sites are using Fingerprintjs2, we need to provide compatibility between the two protocols and make fingerprinting as stable across protocols as possible and easy to use for people who don't know about this issue.

I know it's a tough decision to remove support for HTTP, but I think we should let the user make their own decision.

Valve:
Thanks for your reply @DemonLms .

At this time I cannot merge this PR as it decreased the entropy for the identification process. Starting from v3, this library will not be supported on HTTP websites (it will still work, but we will not be able to guarantee it works correctly). 
Closing, please feel free to reopen with additional comments.

PR #543, opened 2020-09-04
Replace Travis CI with Github Actions, fix the tests

Finesse:


Valve:
@Finesse please squash commits to provide a single, coherent message. 
I am guilty of that myself (not having good commit messages) but as a public project, we should provide nice commit messages.

Finesse:
@Valve Ok. I tried to follow the current repository commit style but I agree that squashed commit look more clean.

PR #544, opened 2020-09-07
Use a more reliable addBehavior check implementation

### LINKED ISSUE #376: addBehaviorKey component could be done more reliable
The `addBehaviorKey` function currently won't work when body isn't defined or was removed.
I suggest to change it from

```
addBehaviorKey: function (keys) {
      // body might not be defined at this point or removed programmatically
      if (!this.options.excludeAddBehavior && document.body && document.body.addBehavior) {
        keys.addPreprocessedComponent({key: 'add_behavior', value: 1})
      }
      return keys
},
```
to

```
addBehaviorKey: function (keys) {
      if (!this.options.excludeAddBehavior && HTMLElement.prototype.addBehavior) {
        keys.addPreprocessedComponent({key: 'add_behavior', value: 1})
      }
      return keys
},
```

which should be more reliable, but i currently have no win machine for testing.

Finesse:
Resolves #376

### DIFF
--- a/fingerprint2.js
+++ b/fingerprint2.js
@@ -466,7 +466,8 @@
     done(hasIndexedDB(options))
   }
   var addBehaviorKey = function (done) {
-    done(!!window.HTMLElement.prototype.addBehavior)
+    // IE8 doesn't have window.HTMLElement and IE9+ doesn't have addBehavior in window.Element
+    done(!!(window.HTMLElement || window.Element).prototype.addBehavior)
   }
   var openDatabaseKey = function (done) {
     done(!!window.openDatabase)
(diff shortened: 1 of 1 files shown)


Finesse:
I've checked in IEs 8–11, Edge 18 and Chrome 85, the result of the component is the same.

Valve:
Thanks @Finesse, please avoid providing any IE9 (or below) compatibility. The latest IE version this lib is compatible with is IE10+. 
We'll drop the support of IE in v3. The v2 is only compatible with IE10+.

Finesse:
@Valve Ok, I couldn't know about this, I've seen multiple signs that say that the library supports IE8. If we support only IE10+, then I'll close all the issues that regard IE8 support. It's a good practice to have a "browser support" section somewhere in documentation so that users know what to expect.

Valve:
@Finesse we have it here: https://docs.fingerprintjs.com/pro/js-agent#supported-browsers
We'll move it into a dedicated page later.

PR #559, opened 2020-10-07
Setup tests in v3

Finesse:
It contains the testing scaffolding and some tests. It doesn't contain all required tests.

The PR will be draft until the pro counterpart is done.


Valve (approved):

.github/workflows/test.yml
```diff
@@ -0,0 +1,26 @@
...
+        uses: actions/checkout@master
+      - name: Install Node packages
+        run: yarn install
+      - name: Build
+        run: yarn build
+      - name: Test on BrowserStack
+        env:
+          # This is in plaintext on purpose. It has no privileged access to anything (this is a free account)
+          # and it allows us to run browserstack tests against PRs.
+          BROWSERSTACK_USERNAME: github13kRcPC
```
Lisenish, line 24: I see your comment but still why not use Github secret here, it's doesn't look very hard.
Maybe someone exceeds our free quota on BrowserStack just because they're lazy to register their own account.

But you have to remove this commit from history in order to fix it now...

src/sources/index.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    it('handles errors', async () => {
+      const sources = {
+        success1: () => 'foo',
+        throwsErrorObject: () => { throw new Error('bar') },
+        throwsErrorString: () => { throw 'baz' },
+        success2: () => 'baq',
+      }
+      const components = await getComponents(sources, undefined, [])
+      expect(Object.keys(components).sort()).toEqual(['success1', 'success2', 'throwsErrorObject', 'throwsErrorString'])
+      expect(components.success1.error).toBeUndefined()
```
Lisenish, line 14: Maybe create assertion helper function, like `expectSucessfulValue(components.success1)`. And vice versa for an error result below

src/sources/index.test.ts
```diff
@@ -0,0 +1,47 @@
+import { getComponents } from './index'
+
+describe('Sources', () => {
+  describe('getComponents', () => {
+    it('handles errors', async () => {
+      const sources = {
```
Lisenish, line 6: Maybe it's a matter of taste but I think tests more readable when you separate arrange-act-assert blocks by a new line.

src/sources/index.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    })
+
+    it('measures duration', async () => {
+      const sources = {
+        instant: () => true,
+        delayedResult: () => new Promise((resolve) => setTimeout(resolve, 50)),
+        delayedError: () => new Promise((_resolve, reject) => setTimeout(() => reject('test'), 50)),
+      }
+      const components = await getComponents(sources, undefined, [])
+      expect(components.instant.duration).toBeLessThan(25)
```
Lisenish, line 32: Why it's 25 here? 
Also maybe it's excessive here but we may mock timeout https://jasmine.github.io/2.0/introduction.html#section-Mocking_the_JavaScript_Timeout_Functions

tests/karma.local.config.js
```diff
@@ -0,0 +1,35 @@
...
+        module: 'commonjs',
+        sourceMap: true,
+      },
+    },
+
+    specReporter: {
+      suppressErrorSummary: true,
+      suppressPassed: true,
+      suppressSkipped: true,
+      // todo: Suppress the summary completely when https://github.com/mlex/karma-spec-reporter/issues/83 is solved
```
Lisenish, line 28: Looks like this repo is not maintained anymore :(

Lisenish (approved):
Good job @Finesse! Only minor comments/suggestions

.github/workflows/test.yml
Finesse: The main reason is to be able to test external PRs automatically; GitHub secrets aren't available in PRs. We can change the token any moment. I had this concert too, I've make a quick research and have found out that other projects keep the keys open too.

src/sources/index.test.ts
Finesse: I think that checking a component status is a very short and clear procedure. Adding a helper function will add a level of abstraction that may increase the cognitive complexity of the code.

src/sources/index.test.ts
Finesse: I agree

src/sources/index.test.ts
Finesse: Because lower values lead to unstable tests that fail occasionally. Timer mocking seems to be an elegant solution, I'll consider it, thank you.

tests/karma.local.config.js
Finesse: Unfortunately. I don't want to spend much time and afford in this testing setup because it isn't battle-checked and there is a chance that it won't be able to run incognito testing. Maybe we will switch to another testing approach.

src/sources/index.test.ts
Finesse: Unfortunately, `jasmine.clock` isn't suitable:

1. It doesn't work well with asynchronous code: https://github.com/jasmine/jasmine/issues/1659
2. It doesn't adjust JavaScript internal clock (that is used to measure duration in this test)

These issues are solvable, but I don't think that this test is worth spending time on looking for a better `jasmine.clock()` alternative.

PR #561, opened 2020-10-09
v3 readme

Finesse:
Some links are missing. They will be added when the pro documentation is ready.


readme.md
```diff
@@ -1,88 +1,183 @@
...
 ></script>
 ```
 
-A simple way to start but not recommended because AdBlock and similar browser extensions often block this URL.
-You should at least upload the script to your server.
+This is a simple way to start but not recommended because AdBlock and other browser extensions can block the script.
+You should at least upload the script file to your server.
 
-### Webpack/Rollup/Browserify
+### Or from NPM and use with Webpack/Rollup/Browserify
```
Valve, line 45: Install from NPM to use with ..

readme.md
```diff
@@ -1,88 +1,183 @@
...
 
 ```js
-import * as FPJS from '@fingerprintjs/fingerprintjs';
+import * as FingerprintJS from '@fingerprintjs/fingerprintjs';
 
 (async () => {
-  const fpjs = await FPJS.load();
-  const { visitorId } = await fpjs.get();
-  console.log(visitorId)
+  // It's good to call this when the application starts
```
Valve, line 57: We recommend to call `load` at application startup.

readme.md
```diff
@@ -1,88 +1,183 @@
...
+  console.log(visitorId);
 })();
 ```
 
-## Version policy
+## Pro version
 
-The library doesn't promise same visitor identifier between any versions
-but tries to keep them same as much as reasonable.
+Upgrade to [Pro version](https://fingerprintjs.com) to get 99.5% accuracy of identification.
```
Valve, line 71: Bold 99.5% and make it orange (our orange color)

readme.md
```diff
@@ -1,88 +1,183 @@
...
+<p align="center">
+  <a href="https://fingerprintjs.com">
+    <img src="resources/pro_screenshot.png" alt="Pro screenshot" width="697px" />
+  </a>
+</p>
 
-## Contribution
+| | Open Source version | Pro version |
+|-----|-------|-------|
+| Browser identification | Basic | Advanced |
```
Valve, line 81: Can we replace this with: Identification accuracy   60%.  99.5%

readme.md
```diff
@@ -1,88 +1,183 @@
...
+| Anonymous user identification | ❌ | ✅ |
```
Valve, line 82: we should remove this as we don't identify users, only browsers, even in the pro

readme.md
```diff
@@ -1,88 +1,183 @@
...
-# Make sure you have Yarn installed
-yarn install
-yarn playground:start # Add '--port 8765' to change the server port
+```js
+{
+  "requestId": "HFMlljrzKEiZmhUNDx7Z",
+  "visitorId": "kHqPGWS1Mj18sZFsP8Wl",
+  "visitorFound": true,
+  "incognito": false,
+  "bot": undefined,
```
Valve, line 98: please replace with an inline example response

readme.md
```diff
@@ -1,88 +1,183 @@
...
+  "ip": "192.65.67.131",
```
Valve, line 99: place immediately before ipLocation

readme.md
```diff
@@ -1,88 +1,183 @@
...
+  "browserName": "Chrome",
+  "browserVersion": "85.0.4183",
+  "os": "Mac OS X",
+  "osVersion": "10.15.6",
+  "device": "Other",
+  "ipLocation": {/* ... */}
```
Valve, line 105: can we have inline lat and lng attributes there ?

readme.md
```diff
@@ -1,88 +1,183 @@
...
 
-To build the playground distribution code (e.g. to upload to a static server), run:
+⏱ [How to upgrade from Open Source to Pro in 30 seconds]()
 
-```bash
-yarn playground:build
-```
+📗 [FingerprintJS Pro documentation](https://dev.fingerprintjs.com)
+
+## Open version reference
```
Valve, line 115: This is confusing, should just call Open-source version reference

readme.md
```diff
@@ -1,88 +1,183 @@
...
+
+The library is shipped in various formats:
+
+- Plain JS
```
Valve, line 119: is it a thing? No better name for it?

readme.md
```diff
@@ -1,88 +1,183 @@
...
 
-The result will appear at `playground/dist`.
+API:
 
-### How to build
+- `FingerprintJS.load({ delayFallback?: number }): Promise<Agent>`
 
-To build the distribution files of the library, run:
+    Builds an instance of Agent and waits a delay required for a proper operation.
+    `delayFallback` is duration (ms) of the fallback for browsers that don't support [requestIdleCallback](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback).
```
Valve, line 144: This should emphasize that this parameter is optional and has a good default which is not recommended to change

readme.md
```diff
@@ -1,88 +1,183 @@
...
+- `agent.get({ debug?: boolean }): Promise<{ visitorId: string, components: {/* ... */} }>`
+
+    Gets the visitor identifier.
+    `debug: true` prints debug messages to the console.
+    `visitorId` is the visitor identifier.
+    `components` is a dictionary of components that have formed the identifier;
+    each value is an object like `{ value: any, duration: number }` in case of success
+    and `{ error: object, duration: number }` in case of an unexpected error during getting the component.
+
+- `FingerprintJS.componentsToCanonicalString(components: object): string`
```
Valve, line 155: why do we need this function to be public? Can we have `FingerprintJS.hashComponents(components: object)` and make this function private?

readme.md
```diff
@@ -1,88 +1,183 @@
...
+
+- `FingerprintJS.getHash(value: string): string`
+
+- `FingerprintJS.componentsToDebugString(components: object): string`
+
+    Converts a dictionary of components (described above) to a human-friendly text.
+
+## Version policy
+
+The library doesn't promise same visitor identifier between any versions
```
Valve, line 169: This library does not guarantee the same visitor identifier between versions, but will try to keep them the same as much as possible.

readme.md
```diff
@@ -1,88 +1,183 @@
...
+but tries to keep them same as much as reasonable.
+
+The documented JS API follows [Semantic Versioning](https://semver.org).
+Using undocumented features is at you own risk.
```
Valve, line 173: Use undocumented features at your own risk.

readme.md
```diff
@@ -1,88 +1,183 @@
...
 
-1. Bump the version. Changing the number in [package.json](package.json) is enough.
-2. Build the project.
-3. See what's will get into the NPM package, make sure it contains the distributive files and no excess files.
-    To see, run `yarn pack`, an archive will appear nearby, open it with any archive browser.
-4. Run
-    ```bash
-    yarn publish --access public # Add '--tag beta' (without the quotes) if you release a beta version
-    ```
+See the [contribution guidelines](contributing.md) to know how to start a playground, test and build.
```
Valve, line 183: to learn how to ..

Valve (changes requested):

readme.md
Finesse: I want “or” to show that this is an alternative way to install, not the second step of installation

readme.md
Finesse: AFAIK markdown doesn’t support text coloring. The only way to color is to put the text as a picture. As the sentence is long, the picture will either be long too or split into lines. In the first case the text will be very small on mobile devices, in the second it will look untidy on desktop.

Also I doubt that making it bold will make it more noticeable. I’d prefer to make this sentence be the section header.

readme.md
Finesse: Do you mean to replace with an example bot object?

readme.md
Finesse: Only them? There are many attributes hidden here

readme.md
Finesse: What would you call it? It can be called:

- Simple `<script>` tag
- HTML script
- IIFE
- Global variable 

It’s not correct to call it “CDM”, because it’s not required to load the script from CDN and UMD can be loaded from CDN too.

readme.md
Finesse: We don’t, actually publishing only a single components-to-hash function is better for us.

readme.md
Valve: ok

readme.md
Valve: then maybe "Alternatively you can install from NPM to use with..."?

Finesse: LGTM

Valve: ok thanks, please update other minor wording changes and let's merge into v3

Finesse: Sure, I'm going to do it on Monday

spalt08:
Can we have a default export with all entities?
```js
import FPJS from '@fingerprintjs/fingerprintjs'
```

Finesse:
> Can we have a default export with all entities?
> 
> ```js
> import FPJS from '@fingerprintjs/fingerprintjs'
> ```

Yes, but it will have no TypeScript types inside because a default export is a single entity and it's impossible to have both types and values in a entity. I'll add a default export that's an object that contains all the documented exported values.

spalt08:
> Yes, but it will have no TypeScript types inside because a default export is a single entity and it's impossible to have both types and values in an entity. I'll add a default export that's an object that contains all the documented exported values.

You can export namespace with:
```js
export as namespace FPJS
```
The AWS package does this, for example

Finesse:
> You can export namespace with:
> 
> ```js
> export as namespace FPJS
> ```
> 
> The AWS package does this, for example

1. The library doesn't export global namespace
2. This syntax is for declaration files only:

    image
3. Users can type `import FPJS, { Agent } from '@fingerprintjs/fingerprintjs` when they need the types. This will be a very rare case, I think.

I've implemented your suggestion. Feel free to improve the typing.

Finesse:
What do you think about centering the pro header?

image

Finesse:
@Valve I've addressed your notes, please take a look

readme.md
```diff
@@ -1,88 +1,197 @@
...
+  </a>
+  <a href="https://www.npmjs.com/package/@fingerprintjs/fingerprintjs">
+    <img src="https://img.shields.io/npm/dt/fingerprintjs2.svg" alt="Total downloads from NPM">
+  </a>
+  <a href="https://www.npmjs.com/package/@fingerprintjs/fingerprintjs">
+    <img src="https://img.shields.io/npm/v/@fingerprintjs/fingerprintjs.svg" alt="Current NPM version">
+  </a>
+</p>
+
+Makes a website visitor identifier from browser fingerprint.
```
Valve, line 18: Makes a website visitor identifier from a browser fingerprint.
or
Makes website visitor identifiers from browser fingerprints.

readme.md
```diff
@@ -1,88 +1,197 @@
...
+Unlike cookies and local storage, fingerprint stays the same in incognito/private mode and after browser data purge.
```
Valve, line 19: .. and even after browser data is purged.

readme.md
```diff
@@ -1,88 +1,197 @@
...
+    const FingerprintJS = require('@fingerprintjs/fingerprintjs');
+    ```
+
+### API
+
+- `FingerprintJS.load({ delayFallback?: number }): Promise<Agent>`
+
+    Builds an instance of Agent and waits a delay required for a proper operation.
+    `delayFallback` is an optional parameter that sets duration (milliseconds) of the fallback for browsers that don't support [requestIdleCallback](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback);
+    it has a good default which isn't recommended to change.
```
Valve, line 161: it as a good default value which we don't recommend to change

readme.md
```diff
@@ -1,88 +1,197 @@
...
+    Gets the visitor identifier.
+    `debug: true` prints debug messages to the console.
+    `visitorId` is the visitor identifier.
+    `components` is a dictionary of components that have formed the identifier;
+    each value is an object like `{ value: any, duration: number }` in case of success
+    and `{ error: object, duration: number }` in case of an unexpected error during getting the component.
+
+- `FingerprintJS.hashComponents(components: object): string`
+
+    Converts a dictionary of components (described above) to a short hash string a.k.a. a visitor identifier.
```
Valve, line 174: .. into a short hash string

readme.md
```diff
@@ -1,88 +1,197 @@
...
+    Designed for extending the library with your own components.
 
-The result will appear at `playground/dist`.
+- `FingerprintJS.componentsToDebugString(components: object): string`
 
-### How to build
+    Converts a dictionary of components (described above) to a human-friendly text.
```
Valve, line 179: converts ... into human-friendly format.

Valve (approved):

readme.md
Finesse: Do you mean to leave or discard the "a.k.a. ..." part?

readme.md
```diff
@@ -1,88 +1,202 @@
...
+    Converts a dictionary of components (described above) into a short hash string a.k.a. a visitor identifier.
+    Designed for extending the library with your own components.
+
+- `FingerprintJS.componentsToDebugString(components: object): string`
+
+    Converts a dictionary of components (described above) into human-friendly format.
+
+## Migration from v2
+
+- [Migration guide](https://github.com/fingerprintjs/fingerprintjs/releases/tag/v3.0.0)
```
Finesse, line 183: I'll put the a migration guide to the releases section of the GitHub repository.

Also I suggest to change the version tag format from `1.2.3` to `v1.2.3` because it's an industry standard and the prefix will help to establish automatic deployment in future. @Valve, what do you think about the format change?

readme.md
```diff
@@ -1,114 +1,202 @@
...
+  // This is the visitor identifier:
+  const visitorId = result.visitorId;
   console.log(visitorId);
 })();
 ```
 
-## Version policy
+📕 [Full documentation](#open-source-version-reference)
+
+## Upgrade to [Pro version](https://fingerprintjs.com) to get 99.5% accuracy of identification
```
Valve, line 71: Upgrade to Pro version to get 99.5% identification accuracy

Valve (approved):

Valve (approved):

readme.md
Valve: ok, let's use v1.2.3, thanks

PR #562, opened 2020-10-12
Make a workaround for the languages component in Chrome 86

Finesse:
Chrome 86 was released and it gives different identifiers in normal vs incognito in certain situations. The difference is in languages array (in normal mode languages from previously used  translations are used, whereas in private mode they are not). This is different from Chrome 85 and earlier, when private mode languages also contained previous translation languages.There is the same difference in other Chromium browsers.

### DIFF
--- a/src/sources/languages.ts
+++ b/src/sources/languages.ts
@@ -12,8 +12,8 @@ export default function getLanguages(): string[][] {
 
   if (Array.isArray(n.languages)) {
     // Starting from Chromium 86, there is only a single value in `navigator.language` in Incognito mode:
-    // the value of `navigator.language`. Wherefore the value is ignored in this browser.
-    if (!isChromium() || !isChromium86OrNewer()) {
+    // the value of `navigator.language`. Therefore the value is ignored in this browser.
+    if (!(isChromium() && isChromium86OrNewer())) {
       result.push(n.languages)
     }
   } else if (typeof n.languages === 'string') {
--- a/src/sources/languages.ts
+++ b/src/sources/languages.ts
@@ -1,3 +1,5 @@
+import { isChromium, isChromium86OrNewer } from '../utils/browser'
+
 const n = navigator
 
 export default function getLanguages(): string[][] {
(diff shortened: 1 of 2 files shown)


src/sources/languages.ts
```diff
@@ -9,7 +11,11 @@ export default function getLanguages(): string[][] {
   }
 
   if (Array.isArray(n.languages)) {
-    result.push(n.languages)
+    // Starting from Chromium 86, there is only a single value in `navigator.language` in Incognito mode:
+    // the value of `navigator.language`. Wherefore the value is ignored in this browser.
```
Valve, line 15: What's `wherefore`?

src/utils/browser.ts
```diff
@@ -85,3 +85,17 @@ export function isGecko(): boolean {
     'CanvasCaptureMediaStream' in w,
   ]) >= 4
 }
+
+/**
+ * Checks whether the browser is based on Chromium version ≥86 without using user-agent.
+ * Doesn't checks that the browser is based on Chromium, there is a separate function for this.
+ */
+export function isChromium86OrNewer(): boolean {
```
Valve, line 93: is it safe to call in browsers, other than Chrome?

Valve:
do we have a normal mode test for this?

src/sources/languages.ts
```diff
@@ -9,7 +11,11 @@ export default function getLanguages(): string[][] {
...
+    if (!isChromium() || !isChromium86OrNewer()) {
```
Valve, line 16: can we invert this condition please? It will be easier to understand if we handle this case in an assertive case (if chrome && newChrome){.

src/sources/languages.ts
Finesse: It's a typo

src/utils/browser.ts
Finesse: It's designed to be called and is called after checking that the browser is Chromium

src/sources/languages.ts
Finesse: I agree, done

PR #565, opened 2020-10-15
V3

Finesse:
Please don't squash the comments, I've squashed some excess commits already.


migrating-v2-v3.md
```diff
@@ -0,0 +1,109 @@
...
+### Bower
+
+Bower support is dropped, choose another installation method.
+
+### NPM or Yarn
+
+Update the package version:
+
+```bash
+npm install @fingerprintjs/fingerprintjs@^3
```
Valve, line 22: why do we need to specify a version explicitly? Will just using `npm install` not work?

Finesse: It will until we release v4. It’s not a big problem actually, just `npm install` will be fine.

PR #570, opened 2020-10-20
Setup linting

Finesse:
- Add a command to check the code with ESLint and Prettier
- Add a command to check that the distributive type declaration file has no errors
- Fix the errors
- Make the checks run automatically in CI


Valve (commented):
LGTM, but we should make sure our OSS rules match our Pro rules.
I see that they match, but I thought we already had pro rules before, but I might be mistaken.

PR #577, opened 2020-11-01
Expand browser support, fix a few bugs

Finesse:
- Support elder browsers like IE 11, Chrome 42 and Firefox 48 (a Promise polyfill is required)
- Handle the case when `new Date().getTimezoneOffset()` returns a string
- fix: The `timezoneOffset` component changes due to DST shifts

I'll need to update the v3 migration guide links after merging this PR.


src/sources/timezone_offset.ts
```diff
@@ -1,3 +1,14 @@
+import { toFloat } from '../utils/data'
+
 export default function getTimezoneOffset(): number {
```
Finesse, line 3: Does `timezoneOffset` even give extra entropy compared to the `timezone` component (the timezone name)? I.e. can there be fingerprints with the same `timezone` and different `timezoneOffset`? If no, then the `timezoneOffset` component can be removed completely.

docs/browser_support.md
```diff
@@ -0,0 +1,83 @@
+# Browser support
```
Finesse, line 1: @Valve What do you think about the document?

Finesse: A preview of the document: https://github.com/fingerprintjs/fingerprintjs/blob/feature/expand-browser-support/docs/browser_support.md

Finesse:
We've decided to review after the release

src/sources/timezone_offset.ts
Valve: Yes, there are different timezones that can coincidentally have the same offset (e.g. Moscow and Dubai)

docs/browser_support.md
Valve: for some reason the link returns 404

src/sources/timezone_offset.ts
Finesse: My question is opposite. Can a single time zone JS value (e.g. `Europe/Moscow`) have different time offsets? It could due to the DST shift, but I've fixed it.

docs/browser_support.md
Finesse: @Valve Because the branch has been deleted. Please try this link: https://github.com/fingerprintjs/fingerprintjs/blob/master/docs/browser_support.md

playground/index.ts
```diff
@@ -1,3 +1,4 @@
+import 'promise-polyfill/src/polyfill'
```
Lisenish, line 1: Have you considered using babel and core-js just to not track manually which polyfills we may need? 
Probably overkill and we fine with just js promise but interested in your reasoning.

Finesse: The documentation says that the library requires a single polyfill to work: a Promise polyfill. Having this single polyfill added manually to the playground is a way to check that nothing else is required.

I used Babel with an [automatic core-js polyfilling](https://babeljs.io/docs/en/babel-preset-env#usebuiltins-usage) in a project. Babel added many excess polyfills, the bundle size was too big. It's not a problem on the playground, I just want to mention.

src/sources/timezone_offset.ts
Valve: I don't think it's possible, otherwise that would be multiple timezones.

Finesse: Ok, thanks. So the only reason to keep this component at the moment is to get entropy in IE.

PR #578, opened 2020-11-02
Add a demo

### LINKED ISSUE #556: Implement v3 open demo
Should show the visitor ID and have a nice design (design will be provided separately).
Do after v3 is released.

Finesse:
Closes #556

The page is made to accomplish a few goals:
1. Increase the chance of getting qualitative bug reports on GitHub by providing an easy way to copy debug information
2. Ease local development of the agent by providing a page that runs the agent every time the agent source code changes
3. Make a quick demonstration of the JS agent

The demo page will be build and deployed automatically when a commit is pushed to the `master` branch.
The demo is and will be available at https://fingerprintjs.github.io/fingerprintjs/.


spalt08 (approved):
Looks good for me

Valve:
Not working in Chrome 42:
image

The playground should have the same level of compatibility as the main product.

Please also make the fonts brighter

Finesse:
@Valve

> Not working in Chrome 42

Fixed

> Please also make the fonts brighter

I've increased the contrast. What do you think?

| Before | After |
|------|------|
| Screen Shot 2020-11-03 at 15 34 12 | Screen Shot 2020-11-03 at 15 35 36 |

playground/index.ts
```diff
@@ -1,25 +1,108 @@
...
+  }
+  document.body.removeChild(textarea)
+}
+
+async function share(text: string) {
+  if (!navigator.share) {
+    alert(`Sharing is unavailable.
+
+Sharing is available in mobile browsers and only on HTTPS websites. ${
+      location.protocol === 'https'
```
borislobanov92, line 95: `location.protocol` return `https:`, so this ternary isn't working correctly for me

playground/index.ts
```diff
@@ -1,25 +1,108 @@
...
+  }
+  contentElement.textContent = content
+  output.appendChild(contentElement)
+}
+
+function initializeDebugButtons(debugText: string) {
+  const copyButton = document.querySelector('#debugCopy')
+  if (copyButton instanceof HTMLButtonElement) {
+    copyButton.disabled = false
+    copyButton.addEventListener('click', async (event) => {
```
borislobanov92, line 60: any reason the event handler is async?

playground/index.ts
```diff
@@ -1,25 +1,108 @@
...
+function copy(text: string) {
+  const textarea = document.createElement('textarea')
+  textarea.value = text
+  document.body.appendChild(textarea)
+  textarea.focus()
+  textarea.select()
+  try {
+    document.execCommand('copy')
+  } catch {
+    // Do nothing in case of a share abort
```
borislobanov92, line 85: looks like a copy-pasted comment from the 'share' method

playground/index.html
```diff
@@ -1,58 +1,146 @@
...
+      box-shadow: 0 0 1.25em rgba(0, 0, 0, 0.2);
+      border-radius: 0.7em;
+      padding: 1em 1.25em;
+      color: #e1e5ea;
+      overflow: auto;
+    }
+    .output pre {
+      padding: 0;
+      margin: 0;
+      font-size: 123%;
```
borislobanov92, line 95: it seems a bit strange that we would use number other than 100% for default and then do 100% for small. why not do 100% for default, 75% for small and 200% for big? I'm probably missing something, so I'd appreciate if you explained your logic.

playground/index.html
```diff
@@ -1,58 +1,146 @@
...
+    .output pre.big {
+      font-size: 200%;
+    }
+    .output pre.small {
+      font-size: 100%;
     }
-    .hint {
-      font-size: 14px;
+    .output .heading {
+      font-size: 75%;
```
borislobanov92, line 107: this font size looks too tiny IMO, is there a reason you decided to use 75% here?

borislobanov92 (changes requested):

playground/index.html
Finesse: The parent font size is `16px`. If a child `pre` font size is set to `100%`, the result font size is `13px` for some reason. This phenomenon persists in different browsers. I want `16px` size for the default so I've set `font-size: 123%`.

playground/index.ts
Finesse: You are right, fixed

playground/index.ts
Finesse: The function was asynchronous at some moment of development, I've forgotten to remove the keyword

playground/index.ts
Finesse: You are right, fixed

playground/index.html
Finesse: I want the headers to be small to not steel the attention, but I agree that 75% is too small. I've increased the font size.

Finesse: I want the headers to be small to not steel attention. But I agree that 75% is too small, I've increased the font size.

playground/index.html
Lisenish: Agree seems strange. Maybe it would be easier to use rem values? 

I also wouldn't use em's for layout values (padding, margins) due to potential cascading issues, but it looks like it would be too much work to change it now and it looks good now, so it's fine.

.github/workflows/demo.yml
```diff
@@ -0,0 +1,30 @@
+# Reference on this file: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions
+# To make it work, go to the repository settings, then Options, GitHub Pages and select `gh-pages` under Source.
+name: Publish the demo
+on:
+  push:
+    branches:
+      - master
+      - feature/556-playground # todo: Remove before merging to master
```
Lisenish, line 8: Don't forget it :)

Lisenish (approved):

playground/index.html
Finesse: @Lisenish `rem` is affected by this phenomenon too. If I set `font-size: 1rem;`, the `pre` (and only `pre`) font size will be `13px`.

PR #579, opened 2020-11-04
Hide the BrowserStack key

Finesse:


Lisenish:
You aware that it's technically still there while we don't purge it from git history, right? (which is a little tricky)

Finesse:
@Lisenish

> You aware that it's technically still there while we don't purge it from git history, right? (which is a little tricky)

The published key doesn’t work anymore, you can try.

Lisenish:
@Finesse Oh, you swapped them, ok. I trust you, just wanted to double-check

PR #584, opened 2020-11-13
Add WebAssembly features detection

kalexey89:
Checklist of features:
- functionReferences - Typed function references and closures (pre-proposal)
- streaming - Instantiates a module directly from a streamed underlying source 
- typeReflection - Type Reflection (--experimental-wasm-type-reflection)
- bigInt - BigInt to WebAssembly i64 integration (--experimental-wasm-bigint)
- bulkMemory - Bulk memory operations (--experimental-wasm-bulk-memory)
- exceptions - exception handling (--experimental-wasm-eh)
- multiValue - Multi values (--experimental-wasm-mv)
- mutableGlobal - Import & export of mutable global (--experimental-wasm-mut-global)
- referenceTypes - Basic reference types "externref" (--experimental-wasm-reftypes)
- saturateConversions - Non-trapping float-to-int conversions (--experimental-wasm-sat-f2i-conversions)
- signExtensions - Zero and sign extensions (--experimental-wasm-se)
- simd - SIMD (--experimental-wasm-simd)
- tailCall - Tail call optiminations (--experimental-wasm-return-call)
- threads - Threads and atomics (--experimental-wasm-threads)

Example:
```javascript
{
  bigInt: true,
  bulkMemory: true,
  exceptions: false,
  functionReferences: false,
  multiValue: true,
  mutableGlobal: true,
  referenceTypes: false,
  saturateConversions: true,
  signExtensions: true,
  simd: false,
  streaming: true,
  tailCall: false,
  threads: true,
  typeReflection: false
}
```


Valve:
Thanks for the PR @kalexey89! We'll review it next week

Finesse:
@kalexey89 What do these features detect: browser or hardware? Will they all be `true` eventually on the same device after several browser updates?

kalexey89:
Hi, @Finesse!

The set of available features depends on JavaScript engine version (so - yes, it depends from browser version) and enabled WebAssembly flags (for example - on "chrome://flags" page, where you may enable experimental features). 

Because WebAssembly is relatively young technology, at now set of available features may change to often. But in the future I think that will be stabilized.

Finesse:
@kalexey89 Thanks. Browser version detection isn't quite suitable for fingerprinting because it's unstable (changes after browser updates) and provides little entropy (most of visitors use the latest version of Blink nowadays and very few change browser flags). Do you know if there are ways to identify hardware or personal settings of ordinary user using WebAssembly?

@Valve, do you agree with my statement?

kalexey89:
@Finesse, I was wondering about determining information about the "hardware" through WASM.
Methods are platform dependent. You will have to load coresponding WASM module depending on the platform.
Plus, it is not yet known whether the browser will give access to such system calls. I think more likely no than yes.

There were also thoughts of "wrapping" the fingerprinting code inside the WASM module, so that it would be more difficult to detect and interfere with its work.

Valve:
It looks like most of these WASM capabilities are only available if you pass certain flags.
These are not useful and we cannot use them as very few users will be enabling those flags. 

The only capability not hidden behind a flag is "function reference". It's unclear if it will provide any additional entropy beyond user-agent, but it looks like it won't.

Another reason why we cannot merge this PR is that it has no tests.

Finesse:
@kalexey89 

> Methods are platform dependent. You will have to load coresponding WASM module depending on the platform.
Plus, it is not yet known whether the browser will give access to such system calls. I think more likely no than yes.

What methods do you mean? If they seem to be possible, then we shall try them.

> There were also thoughts of "wrapping" the fingerprinting code inside the WASM module, so that it would be more difficult to detect and interfere with its work.

Source code of the library is open, so there is no point of hiding it. Cann you tell a use-case where the JS code should be hidden even though its source is available publicly?

PR #585, opened 2020-11-16
Improve the available screen resolution component

### LINKED ISSUE #568: Available screen resolution changes when in fullscreen mode on MacOS
When I'm entering full screen mode in MacOS Chrome the fingerprint changes because of the available screen resolution.

Finesse:
Resolves #568

- The component is disabled in browsers where it's unstable or gives no information. It's enabled in: Chrome ≤ 83 on Windows, Firefox on Windows, IE, pre-chromium Edge, Chrome 57–78 on macOS, Safari ≤11 on macOS. Approximate share of such users is 9.3% and it decreases. See raw research data at https://github.com/fingerprintjs/fingerprintjs/issues/568#issuecomment-722272726.
- Top-left corner position of available screen is added.
- The dimensions are rounded by 10 pixels to mitigate small random changes of the position (e.g. 1900x1440 → 1900x1439, the reason and steps to reproduce are unknown).

This PR will change every visitor identifier.


spyjo:
Hello,
How many users are using fullscreen mode ? I guess less than 1%
If I understand well, you want to replace a component which works for 99% of users, to a component which work for only 9.3% of users and decreasing, and moreover changing all visitor ID ? Please correct me if I'm wrong.

Valve:
@Finesse I think this change and its consequences are more nuanced and we shouldn't hurry with merging it. 
Let's discuss it first.

Finesse:
@spyjo Only 40% of visitors use desktop devices. This component gives no information on 99.5% of mobile devices and spoils fingerprint on the rest 0.5% of mobile devices so removing it for mobile devices will only increase accuracy a bit.

Fullscreen is very common on macOS, it's very convenient, but the macOS share is pretty small (about 3% of desktops). I also suppose that Windows users use UI fullscreen (<kbd>F11</kbd>) very rarely too, and such fullscreen doesn't change the component result. But [programmatic fullscreen mode](https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullscreen) does change the result in Blink (Chromium-like browsers) and programmatic fullscreen mode is used often on some websites such as YouTube. There are 2 options to handle this case: either ask library users to not run the library in programmatic fullscreen mode or disable the component in Blink on Windows.

Finesse:
I've tried a few tricks to get real available screen resolution in programmatic fullscreen mode on Windows including running the code in an iframe and in a separate window. They all have failed. If you have an idea to solve it, please share.

I think that UI fullscreen is an issue on macOS and it should be considered. I think that programmatic fullscreen is an issue on all platforms and should be considered even though fingerprint is very unlikely to be taken in such situation.

Browsers can be split into following groups:

1. Available screen resolution is always stable and gives entropy. It's about 9% of all visitors and it decreases. Browsers:
    - Chrome ≤83 on Windows
    - Firefox on Windows
    - IE
    - pre-chromium Edge
    - Chrome 57–78 on macOS
    - Safari ≤11 on macOS
2. Available screen resolution is stable and gives entropy unless OS fullscreen is used. It's about 4.2% of all visitors. Browsers:
    - Chrome ≤56 on macOS
    - Safari ≥12 on macOS
3. Available screen resolution is stable and gives entropy unless programmatic fullscreen is used. It's about 23% of all visitors and it increases. Browsers:
    - Chrome ≥84 on Windows
4. Available screen resolution is stable and gives entropy unless any fullscreen is used. It's about 9% of all visitors and it increases. Browsers:
    - Chrome OS
    - Firefox on macOS
    - Chrome ≥79 on macOS
5. Available screen resolution is unstable at all (depends on the device orientation and not only swaps the numbers). It's about 0.2% of all visitors and it decreases. Browsers:
    - Android Firefox ≤68
    - All browsers on iOS 9
6. Available screen resolution is always equal to screen resolution and therefore gives no entropy. It's about 55%. Browsers: all other mobile browsers.

Let's say that 40% of macOS and Chrome OS users toggle OS fullscreen, and 0.5% of all users have their fingerprint calculated during programmatic fullscreen. If so, then the component precision at `master` branch at the moment is:

| Useful entropy | False negative (unexpected change of fingerprint) | False positive (lost chance to distinguish visitors) | Evaded false negative | No entropy |
|----|-----|----|----|----|
| 9% + 4.2% * 0.6 + 23% * 0.995 + 9% * 0.6 * 0.995 ≈ **39%** | 4.2% * 0.4 + 23% * 0.005 + 9% * (1 - 0.6 * 0.995) + 0.2% ≈ **6%** | **0%** | **0%** | **55%** |

_(the tables show shares of visitors who's prone to specific situations, not shares of results of visitor identifications)_

The merge request code state at commit 21b7fdca0e1fca710a49ad07e677655a9afbcaf9 discards the component value in all the browser groups except the 1st, so the precision is:

| Useful entropy | False negative | False positive | Evaded false negative | No entropy |
|----|-----|----|----|----|
| **9%** | **0%** | 4.2% * 0.6 + 23% * 0.995 + 9% * 0.6 * 0.995 ≈ **30%** | 4.2% * 0.4 + 23% * 0.005 + 9% * (1 - 0.6 * 0.995) + 0.2% ≈ **6%** | **55%** |

So the error rate has increased from 6% to 30% and the error type has switched from false negative to false positive. False positive is more preferable because it can be compensated by other components and false negative can't. But I think the change is too serious. My bad that I hadn't made these estimations before implementing the solution.

Luckily, I've found a way to make the component be stable despite programmatic fullscreen. The library can get and cache available screen resolution before even `FingerprintJS.load()` is called and/or [turn off programmatic fullscreen](https://developer.mozilla.org/en-US/docs/Web/API/Document/exitFullscreen) before running the component. It's a radical solution, but it will be used in _super rare_ cases. Using this solution and enabling the component for the 3rd group will change the precision to:

| Useful entropy | False negative | False positive | Evaded false negative | No entropy |
|----|-----|----|----|----|
| 9% + 23% ≈ **32%** | **0%** | 4.2% * 0.6 + 9% * 0.6 ≈ **8%** | 4.2% * 0.4 + 9% * 0.4 + 0.2% ≈ **5%** | **55%** |

I.e. the error rate will increase from 6% to 8%, but the errors type will switch from false negative to false positive. This is a good deal, I think.

@Valve @spyjo What do you think?

src/utils/data.ts
```diff
@@ -44,6 +44,21 @@ export function toFloat(value: unknown): number {
...
+export function replaceNaN<T, U>(value: T, replacement: U): T | U {
+  return typeof value === 'number' && isNaN(value) ? replacement : value
+}
+
 export function countTruthy(values: unknown[]): number {
   return values.reduce<number>((sum, value) => sum + (value ? 1 : 0), 0)
 }
+
+export function round(value: number, base = 1): number {
+  if (Math.abs(base) >= 1) {
```
spalt08, line 56: you can also do here `value - value % base`. Should work for both cases

src/sources/available_screen_position.ts
```diff
@@ -0,0 +1,39 @@
...
+  if (!isUseful()) {
+    return undefined
+  }
+
+  // Some browsers return screen resolution as strings, e.g. "1200", instead of a number, e.g. 1200.
+  // I suspect it's done by certain plugins that randomize browser properties to prevent fingerprinting.
+  //
+  // Sometimes the resolution changes a bit for unknown reason, e.g. 1900x1440 → 1900x1439,
+  // the rounding is used to mitigate this difference.
+  const prepareValue = (value: unknown) => replaceNaN(round(toFloat(value), roundingPrecision), null)
```
spalt08, line 18: just FYI, you can also use `+value` for string to integer conversion, which is faster

spalt08 (approved):

Valve:
In my experience a false positive is a far more serious type of error than a false negative. Usually it's fine to occasionally generate multiple visitorID values per single browser, but it's a worse problem to identify multiple unrelated visitors as one visitor. For this reason I think this proposal shouldn't be implemented. 
As a related thing: it also doesn't handle the position of the system toolbar (left, bottom or right), which can be detected with `avail` values. 
Ideally we need:
1. Detect the system toolbar position and size.
2. Ignore available resolution. 

If we can do the first item consistently somehow, we should do it.

src/sources/available_screen_position.ts
```diff
@@ -0,0 +1,39 @@
+import { replaceNaN, round, toFloat } from '../utils/data'
+import * as browser from '../utils/browser'
+
+type Position = [left: number | null, top: number | null, width: number | null, height: number | null]
```
Valve, line 4: this structure is usually called a Rect.
(e.g. Rect {x: int, y: int, width: int, height: int})
Example: https://docs.microsoft.com/en-us/dotnet/api/system.windows.rect?view=net-5.0

It's not really important, but I noticed that because it didn't really look like a position to me.

tbadlov:
This is also an issue if you attach a secondary monitor to your screen resolution values.
We ended up just removing the components for now.

```
Firefox Native Screen
version: 3.0.3
userAgent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:83.0) Gecko/20100101 Firefox/83.0
getOptions: {
  "debug": true
}
visitorId: d6e7e6f8c195450bcfb80b5d83bee01b
components: {
  "osCpu": {
    "value": "Intel Mac OS X 10.15",
    "duration": 0
  },
  "languages": {
    "value": [
      [
        "en-US"
      ],
      [
        "en-US",
        "en"
      ]
    ],
    "duration": 0
  },
  "colorDepth": {
    "value": 24,
    "duration": 0
  },
  "deviceMemory": {
    "duration": 0
  },
  "screenResolution": {
    "value": [
      900,
      1440
    ],
    "duration": 1
  },
  "availableScreenResolution": {
    "value": [
      877,
      1440
    ],
    "duration": 0
  },

```
```
Firefox - Attached Monitor
version: 3.0.3
userAgent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:83.0) Gecko/20100101 Firefox/83.0
getOptions: {
  "debug": true
}
visitorId: f3a4175eb8112a116b637ef768477ffd
components: {
  "osCpu": {
    "value": "Intel Mac OS X 10.15",
    "duration": 0
  },
  "languages": {
    "value": [
      [
        "en-US"
      ],
      [
        "en-US",
        "en"
      ]
    ],
    "duration": 1
  },
  "colorDepth": {
    "value": 24,
    "duration": 0
  },
  "deviceMemory": {
    "duration": 0
  },
  "screenResolution": {
    "value": [
      1920,
      1080
    ],
    "duration": 0
  },
  "availableScreenResolution": {
    "value": [
      1920,
      1057
    ],
    "duration": 0
  },
```

src/sources/available_screen_position.ts
Finesse: `Rect` is "what it is" and `Position` is "what it means" (a higher level abstraction). I don't mind calling it `Rect`. But I think I'll remove this type soon to implement a better solution.

Finesse:
> In my experience a false positive is a far more serious type of error than a false negative. Usually it's fine to occasionally generate multiple visitorID values per single browser, but it's a worse problem to identify multiple unrelated visitors as one visitor. For this reason I think this proposal shouldn't be implemented.

@Valve Thanks for your opinion. Ok, I won't exclude the component. I'll implement another solution to improve its accuracy without losing the entropy: exit programmatic fullscreen and watch the available screen resolution before `get()` is called.

> As a related thing: it also doesn't handle the position of the system toolbar (left, bottom or right), which can be detected with `avail` values.

It does. You can see that all 4 measurements are taken at https://github.com/fingerprintjs/fingerprintjs/pull/585/files#diff-6411a5d12f7f63f6085e73468c1d6c96b5505da2b1b2404ff1493f9e50a904f1R21

> Ideally we need:
> 1. Detect the system toolbar position and size.
> 2. Ignore available resolution.
>
> If we can do the first item consistently somehow, we should do it.

I think it's a good idea. I'm going to make the component do measure toolbar sizes at all 4 sides. It's more reliable because toolbar sizes are unlikely to change when screen is rotated.

> This is also an issue if you attach a secondary monitor to your screen resolution values.
We ended up just removing the components for now.

@tbadlov Yes, it's a known problem, there is nothing we can do with it. Do you have an idea how to mitigate this? Consider that a visitor can have many identifiers due to various changes, you can either exclude unstable components or improve your system to handle multiple identifires.

Finesse:
I've found a possible reason for the available screen resolution to change a bit. When macOS Dock contains too many icons to fit them all in one row, it shrinks to fit more, i.e. the toolbar size is decreased. This changes the values of `screen.avail...`.

This is message is for historical purposes. This feature will be mitigated by the rounding.

Finesse:
I've implemented the solution that I mentioned at https://github.com/fingerprintjs/fingerprintjs/pull/585#issuecomment-731553179. See [the latest commit](https://github.com/fingerprintjs/fingerprintjs/pull/585/commits/11013d3d5c9b63f793c36f6504373023f9424068) for the change.

@Valve @spyjo What do you think?

readme.md
```diff
@@ -26,15 +26,17 @@ Unlike cookies and local storage, fingerprint stays the same in incognito/privat
...
+    // Get a visitor identifier when you need it.
+    fpPromise
+      .then(fp => fp.get)
+      .then(result => {
         // This is the visitor identifier:
         const visitorId = result.visitorId;
         console.log(visitorId);
       });
-    });
   }
```
Finesse, line 40: Having a big timespan between starting agent and getting an identifier increases the chance of getting the real screen frame size (the frame between screen resolution and available screen resolution).

Valve:
Another idea we can explore is: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/device-height

Finesse:
@Valve I've created a separate branch for commits that change fingerprint: `next-sources`. Please approve this PR and I'll merge it into the branch.

Valve (approved):

swilliams-a3digital:
My available resolution changed 4 times in 1 day on chrome on mac OS and I NEVER entered full screen mode.  It just randomly changes by 10 to 15 pixels throughout the dday.

Valve:
@swilliams-a3digital looks like an edge case to me. Please feel free to exclude it using the documentation link that @Finesse provided.

Valve:
@swilliams-a3digital I'm curious, do you have your dock fixed with dynamic sizing?

swilliams-a3digital:
I am using a mac book Pro.  I don't have a monitor attached, just using the laptop screen.  Display settings are set to
Resoulution : Default for Display.
I pretty much have the out of the box setup for a macbook without monitor attached.

Valve:
I meant more your dock settings or your other software settings. Please ignore this settings if it's a concern. For extended discussions, I would like to welcome you to github discussions

Finesse:
@swilliams-a3digital Does your Dock size change within a day? MacOS Dock shrinks when there are too many icons in it to fit, the shrinking changes the available screen resolution. Please send us a couple of full screenshots: one when `availableScreenResolution` has one value and one when `availableScreenResolution` has another value.

swilliams-a3digital:
Here are two screen shots:
finger-print-res-1792, 1017
finger-print-res-1792, 1120

It's the same screen, but one reports resolution of 1017 and the other 1120, and this leads to making it look like a different fingerprint.

Finesse:
@swilliams-a3digital Thanks for the screenshots. Your screenshots have different resolutions. You either use different monitors (the fingerprint will be different for different monitors) or have cropped the screenshots (we need full screen screenshots to investigate the issue).

PR #597, opened 2020-12-20
Added guide content to support older browsers

### LINKED ISSUE #596: Syntax for Old browsers
#597 
We recently migrated from version 2 to version 3. There were several issues at this time, but the biggest issue was that IE11 did not support Arrow Function as follows, so a syntax error occurred.

For this, I suggest adding content to browser_support.md as follows .(polyfill is still needed)

-----------
## Syntax for Old browsers
Older browsers like IE11 do not support arrow functions (=>). To support older browsers, it should be used in the following form
```diff
- FingerprintJS.load().then(fp => {
+ FingerprintJS.load().then(function (fp) {
-fp.get().then(result => {
+ fp.get().then(function (result) {
```

songmw90:
Closes #596

### DIFF
--- a/docs/browser_support.md
+++ b/docs/browser_support.md
@@ -81,3 +81,12 @@ Examples for various installation methods:
 
       // Start loading FingerprintJS here
     ```
+## Syntax for Old browsers
+
+Older browsers like IE11 do not support arrow functions (=>). To support older browsers, it should be used in the following form
+```diff
+- FingerprintJS.load().then(fp => {
++ FingerprintJS.load().then(function (fp) {
+-fp.get().then(result => {
++ fp.get().then(function (result) {
+```
--- a/docs/browser_support.md
+++ b/docs/browser_support.md
@@ -21,6 +21,7 @@ At the moment, the browsers are:
 Other browsers will probably also work, but we don't guarantee it.
 
 ## Old browsers requirements
+### Polyfills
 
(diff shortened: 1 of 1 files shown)


docs/browser_support.md
```diff
@@ -81,3 +81,12 @@ Examples for various installation methods:
 
       // Start loading FingerprintJS here
     ```
+## Syntax for Old browsers
```
Finesse, line 84: This is a requirement for all browsers. So please make the document have the following structure:

- Old browsers requirements
    - Polyfills
    - Code syntax

docs/browser_support.md
```diff
@@ -81,3 +81,12 @@ Examples for various installation methods:
...
+
+Older browsers like IE11 do not support arrow functions (=>). To support older browsers, it should be used in the following form
+```diff
+- FingerprintJS.load().then(fp => {
++ FingerprintJS.load().then(function (fp) {
+-fp.get().then(result => {
++ fp.get().then(function (result) {
```
Finesse, line 91: Please make the snippet be level be using the same horizontal spacing in all the lines.

Finesse (changes requested):
Thank you for your suggestion, this is a good point. But there are a couple minor issues.

Finesse (approved):
Thank you 👍

Issue #599, opened 2021-01-06
Some cookies are misusing the recommended “SameSite“ attribute


iipajcii:
### Scenario
- What's the scenario, what happens and what did you expect to happen?
I'm making an application using the npm install version of this package and I get two warnings saying:
> Cookie “cookietest” will be soon rejected because it has the “SameSite” attribute set to “None” or an invalid value, without the “secure” attribute. To know more about the “SameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite

- What device and browser are you using?
I'm using Firefox version 82.0.3

- What version of fingerprintjs are you using? (Bug reports not applicable to fingerprintjs master are subject to be closed without comment.)
According the package.json file the current version of fingerprintjs is 3.0.4

### Fingerprint data
Here is an image of the error in the console
![Screenshot from 2021-01-05 21-23-48](https://user-images.githubusercontent.com/57077771/103724138-aadef700-4fa1-11eb-975f-edb2c452c0ca.png)

PR #601, opened 2021-01-06
Fix cookie 'SameSite' attribute warning in Firefox (#599)

iipajcii:
This fix adds the SameSource Attribute and the Secure Attribute to the cookie test.
SameSource is set to 'Strict' so that only the website that the cookie will have access to it, making it a first-party cookie according to [Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite).

### DIFF
--- a/src/sources/cookies_enabled.ts
+++ b/src/sources/cookies_enabled.ts
@@ -18,10 +18,10 @@ export default function areCookiesEnabled(): boolean {
   // or in sandboxed iframes (depending on flags/context)
   try {
     // Create cookie
-    d.cookie = 'cookietest=1'
+    d.cookie = 'cookietest=1; SameSite=Strict; Secure;'
     const result = d.cookie.indexOf('cookietest=') !== -1
     // Delete cookie
-    d.cookie = 'cookietest=1; expires=Thu, 01-Jan-1970 00:00:01 GMT'
+    d.cookie = 'cookietest=1; SameSite=Strict; Secure; expires=Thu, 01-Jan-1970 00:00:01 GMT'
     return result
   } catch (e) {
     return false
--- a/src/sources/cookies_enabled.ts
+++ b/src/sources/cookies_enabled.ts
@@ -18,10 +18,10 @@ export default function areCookiesEnabled(): boolean {
   // or in sandboxed iframes (depending on flags/context)
   try {
     // Create cookie
-    d.cookie = 'cookietest=1; SameSite=Strict; Secure;'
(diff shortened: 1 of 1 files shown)


Finesse:
@iipajcii, thank you for your contribution. You've added both `SameSize` and `Secure` attributes, while Firefox requires either `Secure` or `SameSite`. Looks like your change makes the `cookieEnabled` component always return `false` on `http://` websites, and this can be avoided. Have you checked it?

iipajcii:
> @iipajcii, thank you for your contribution. You've added both `SameSize` and `Secure` attributes, while Firefox requires either `Secure` or `SameSite`. Looks like your change makes the `cookieEnabled` component always return `false` on `http://` websites, and this can be avoided. Have you checked it?

No I have not tried it on `https`, my development environment uses `http'. I could leave `SameSite` set to `Strict` and leave out `Secure`, however I'm not sure if the maintainers intend for this script to only be used in one site, or if it's okay for third party sites to use it.

If third-parties are to have access and if it needs to work over `https` I think setting `SameSite` to `Lax` would work

Finesse:
@iipajcii

> No I have not tried it on `https`, my development environment uses `http`.

I meant you to try HTTP usage, not HTTPS. You've set `Secure` flag, it means that the cookie will be available only on HTTPS pages and therefore the source will always return `false` on HTTP pages. Have you checked it? You've said that you've tried it on an HTTP page; have you tried to disable/enable cookies in browser settings and see if the source returned value changes?

> I could leave `SameSite` set to `Strict` and leave out `Secure`, however I'm not sure if the maintainers intend for this script to only be used in one site, or if it's okay for third party sites to use it. If third-parties are to have access and it needs to work over `https` I think setting `SameSite` to `Lax` would work

The FingerprintJS library is intended to be used on any website, both on HTTP and HTTPS. I think any `SameSite` value other than `None` will work because the cookie doesn't participate in HTTP requests because it's removed right after creation.

iipajcii:
> The FingerprintJS library is intended to be used on any website, both on HTTP and HTTPS. I think any `SameSite` value other than `None` will work because the cookie doesn't participate in HTTP requests because it's removed right after creation

I understand your point, since it's only a test to see if cookies work I've just updated my branch and removed `Secure` and now it only sets `SameSite` to `Strict`.

You should see the change in the 'Files Changed' Section now.

Finesse:
All good, but the changes must go to the `master` branch instead of the `next-sources` branch

PR #627, opened 2021-03-11
sources: local storage: name file by functionality

wiese:
There is no indication in the history that the former name, location_storage.ts, was chosen intentionally but it is quite confusing so let's rename it according to the functionality it provides.


Finesse:
Nice catch, thank you!

PR #666, opened 2021-06-04
Make the entropy sources run in the "load" phase

Finesse:
In order to make agent's `get` function run quickly when some time passes between calling `load` and `get`.

The new signal mechanism description: https://github.com/fingerprintjs/fingerprintjs/blob/feature/load-signals/contributing.md#how-to-make-an-entropy-source


contributing.md
```diff
@@ -94,6 +94,118 @@ To check that the package is compatible with server side rendering, build the pr
...
+  return finalData // Equivalent to: return () => finalData
+}
+```
+
+In fact, most entropy sources don't require a "get" phase.
+The "get" phase is required if the component can change after completing the load phase.
+
+In order for agent to measure the entropy source execution duration correctly,
+the "load" phase shouldn't run in background (after the source function returns).
+On the other hand, in order not to detain the whole agent, the "load" phase must contain only the necessary actions.
```
Valve, line 158: `detain` is unclear, perhaps you meant "block"?

contributing.md
```diff
@@ -94,6 +94,118 @@ To check that the package is compatible with server side rendering, build the pr
...
+Example:
+
+```ts
+async function entropySource() {
+  let result: Record<string, unknown> | undefined
+  const dataPromise = doLongAction().then((value) => result = value) // Simplified. Error handling is required.
+
+  // Apply the minimal required timeout
```
Valve, line 166: This comment is unclear

contributing.md
Finesse: Why is it unclear? Is it a wrong context for the word?

Valve: detain is when a police detains a criminal, it's like "arrest".

contributing.md
Finesse: Does it look good now?

contributing.md
```diff
@@ -94,6 +94,117 @@ To check that the package is compatible with server side rendering, build the pr
...
 
+### How to make an entropy source
+
+Entropy source is a function that gets a piece of data about the browser a.k.a. a fingerprint component.
+Entropy sources are located in the [src/sources](src/sources) directory.
+Entropy component must be a simple JS value that can be JSON-encoded.
+
+Entropy source runs in 2 stages: "load" and "get":
+- "Load" runs once when the agent's `load` function is called.
+    It must do as much work of the source as possible to make the "get" phase as fast as possible.
```
makma, line 105:
```suggestion
    It must do as much work on the source as possible to make the "get" phase as fast as possible.
```

docs/api.md
```diff
@@ -65,18 +65,18 @@ const FingerprintJS = require('@fingerprintjs/fingerprintjs')
...
+#### `FingerprintJS.load({ delayFallback?: number, debug?: boolean }): Promise<Agent>`
 
 Builds an instance of Agent and waits a delay required for a proper operation.
 We recommend calling it as soon as possible.
 `delayFallback` is an optional parameter that sets duration (milliseconds) of the fallback for browsers that don't support [requestIdleCallback](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback);
 it has a good default value which we don't recommend to change.
+`debug: true` prints debug messages to the console.
 
-#### `agent.get({ debug?: boolean }): Promise<object>`
+#### `agent.get(): Promise<object>`
```
makma, line 76: Are we going to release a new major version? From my point of view, [according to the version policy](https://github.com/fingerprintjs/fingerprintjs/blob/master/docs/version_policy.md), this is a breaking change, isn't it? Or do I miss anything?

src/agent.test.ts
```diff
@@ -22,9 +23,9 @@ describe('Agent', () => {
     }
   })
 
-  it('watches screen frame before calling `get()`', async () => {
+  it('loads sources when created', async () => {
+    // Checking by checking whether the screen frame is watched
```
makma, line 27:
```suggestion
    // Checking whether the screen frame is watched
```

docs/api.md
Finesse: It's not a breaking change because the parameter isn't actually removed. The parameter is [marked as deprecated](https://github.com/fingerprintjs/fingerprintjs/pull/666/files#diff-49cc4f0909411e6de89a20d5a97f81c481546b48973288ba70aeb002b0dd4138R32). The old code like this will work, but in a slightly other way:

```js
const fp = await fpPromise
const result = await fp.get({ debug: true })
console.log(result.visitorId)
```

So nothing will break after the update, therefore it's ok to keep the major version.

src/agent.test.ts
Finesse: The full sentence is "Checking whether agent loads entropy sources when created by checking whether the screen frame is watched". How would you write it briefly?

makma: oh, got it, ty

docs/api.md
makma: I see it now, looks good, thank you for the clarification.

spalt08 (approved):
LGTM, but tests are failing
So I approve in advance

PR #686, opened 2021-07-22
Fix: use iframe.contentDocument instead

angelo-yu-ND:
Sometimes iframe.contentWindow is loaded while iframe.contentWindow.document is still null, and cause an exception

### DIFF
--- a/src/utils/dom.ts
+++ b/src/utils/dom.ts
@@ -47,6 +47,7 @@ export async function withIframe<T>(
       // See https://github.com/fingerprintjs/fingerprintjs/issues/645
       const checkReadyState = () => {
         // Make sure iframe.contentWindow and iframe.contentWindow.document are both loaded
+        // The contentWindow.document can miss in JSDOM (https://github.com/jsdom/jsdom).
         if (iframe.contentWindow?.document?.readyState === 'complete') {
           resolve()
         } else {
--- a/src/utils/dom.ts
+++ b/src/utils/dom.ts
@@ -46,6 +46,7 @@ export async function withIframe<T>(
       // This code checks for the loading state manually.
       // See https://github.com/fingerprintjs/fingerprintjs/issues/645
       const checkReadyState = () => {
+        // Make sure iframe.contentWindow and iframe.contentWindow.document are both loaded
         if (iframe.contentWindow?.document?.readyState === 'complete') {
           resolve()
         } else {
--- a/src/utils/dom.ts
+++ b/src/utils/dom.ts
@@ -46,7 +46,7 @@ export async function withIframe<T>(
       // This code checks for the loading state manually.
       // See https://github.com/fingerprintjs/fingerprintjs/issues/645
       const checkReadyState = () => {
-        if (iframe.contentWindow?.document.readyState === 'complete') {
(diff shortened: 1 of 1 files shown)


Finesse:
Hi @angelo-yu-ND, thank you for the contribution. Could you please solve a couple of issues:

1. How to reproduce the error?
2. The TypeScript check of the code fails: `src/utils/dom.ts(62,33): error TS2345: Argument of type 'Window | null' is not assignable to parameter of type 'Window'.`. The line 58 is a [type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types) for the line 62, your change breaks the type guard. So, you can try to replace `iframe.contentDocument?.body` with `iframe.contentWindow?.document?.body`. 
3. Add a small comment above the changed line that mentions the error. It will help preventing regressions.
4. Make the same change to all the code parts that use `iframe.contentWindow?.document`.

Finesse (changes requested):
Hi @angelo-yu-ND, thank you for the contribution. Could you please solve a couple of issues:

1. How to reproduce the error?
2. The TypeScript check of the code fails: `src/utils/dom.ts(62,33): error TS2345: Argument of type 'Window | null' is not assignable to parameter of type 'Window'.`. The line 58 is a [type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types) for the line 62, your change breaks the type guard. So, you can try to replace `iframe.contentDocument?.body` with `iframe.contentWindow?.document?.body`. 
3. Add a small comment that mentions the error above the changed line. It will help preventing regressions.
4. Make the same change to the other code parts that use `iframe.contentWindow?.document`. It's only line 49, as far as I can see.

angelo-yu-ND:
Hi @Finesse , the issue happens frequently while we are writing `jest` tests for React component loading `fingerprintjs`:
```
        at listOnTimeout (internal/timers.js:554:17)
        at processTimers (internal/timers.js:497:7) TypeError: Cannot read property 'readyState' of undefined
        at checkReadyState (/.../frontend/node_modules/new-fingerprint/dist/fp.cjs.js:946:115)
        at Timeout.callback [as _onTimeout] (/.../frontend/node_modules/jest-environment-jsdom/node_modules/jsdom/lib/jsdom/browser/Window.js:643:19)
        ......
```

Thanks for the review!

src/utils/dom.ts
```diff
@@ -46,7 +46,8 @@ export async function withIframe<T>(
       // This code checks for the loading state manually.
       // See https://github.com/fingerprintjs/fingerprintjs/issues/645
       const checkReadyState = () => {
-        if (iframe.contentWindow?.document.readyState === 'complete') {
+        // Make sure iframe.contentWindow and iframe.contentWindow.document are both loaded
```
Finesse, line 49: According to your error stack trace, this situation can happen in [JSDOM](https://github.com/jsdom/jsdom). Could you please edit the comment so that it says that the `contentWindow.document` can miss in JSDOM.

Finesse (changes requested):

Finesse (approved):

Finesse:
Thank you for the contribution!

PR #716, opened 2021-12-04
Use setInterval instead of recursively calling setTimeout

lightmare:
Today I noticed Slack.com constantly hogging cpu (in Firefox 94). After some digging in profiler and through obfuscated/minified js, I landed on this function:
https://github.com/fingerprintjs/fingerprintjs/blob/e0f1771bd8e87535e99bb632b6315c55fd8dd202/src/utils/dom.ts#L48-L56
which spends ~20ms per second in `setTimeout`. Should be using `setInterval` instead.

Then looking through the surrounding code I saw more loops repeatedly calling `setTimeout` while waiting for a condition. So I replaced those with `setInterval` as well.

### DIFF
--- a/src/sources/dom_blockers.ts
+++ b/src/sources/dom_blockers.ts
@@ -1,7 +1,7 @@
 import { isAndroid, isWebKit } from '../utils/browser'
 import { selectorToElement } from '../utils/dom'
 import { countTruthy } from '../utils/data'
-import { wait } from '../utils/async'
+import { waitUntil } from '../utils/async'
 
 /**
  * Only single element selector are supported (no operators like space, +, >, etc).
@@ -326,10 +326,8 @@ export async function getBlockedSelectors<T extends string>(selectors: readonly 
   }
 
   // document.body can be null while the page is loading
-  while (!d.body) {
-    await wait(50)
-  }
-  d.body.appendChild(root)
+  const body = d.body || (await waitUntil(() => d.body, 50))
+  body.appendChild(root)
 
   try {
     // Then check which of the elements are blocked
--- a/src/utils/async.ts
+++ b/src/utils/async.ts
@@ -4,6 +4,24 @@ export function wait<T = void>(durationMs: number, resolveWith?: T): Promise<T> 
   return new Promise((resolve) => setTimeout(resolve, durationMs, resolveWith))
 }
 
+export function waitUntil<T>(condition: () => T, intervalMs: number): Promise<T> {
+  return new Promise((resolve, reject) => {
+    const checkCondition = () => {
+      try {
+        const v = condition()
+        if (v) {
+          resolve(v)
+          clearInterval(timerId)
+        }
+      } catch (e) {
+        reject(e)
+        clearInterval(timerId)
+      }
+    }
+    const timerId = setInterval(checkCondition, intervalMs)
+  })
+}
+
 export function requestIdleCallbackIfAvailable(fallbackTimeout: number, deadlineTimeout = Infinity): Promise<void> {
   const { requestIdleCallback } = window
   if (requestIdleCallback) {
--- a/src/utils/dom.ts
+++ b/src/utils/dom.ts
@@ -1,4 +1,4 @@
-import { wait } from './async'
+import { waitUntil } from './async'
 import { parseSimpleCssSelector } from './data'
 
 /**
@@ -19,8 +19,8 @@ export async function withIframe<T>(
   const d = document
 
   // document.body can be null while the page is loading
-  while (!d.body) {
-    await wait(domPollInterval)
+  if (!d.body) {
+    await waitUntil(() => d.body, domPollInterval)
   }
 
   const iframe = d.createElement('iframe')
@@ -56,11 +56,12 @@ export async function withIframe<T>(
       const timerId = setInterval(checkReadyState, 10)
     })
 
-    while (!iframe.contentWindow?.document?.body) {
-      await wait(domPollInterval)
+    const haveContentBody = () => !!iframe.contentWindow?.document?.body
+    if (!haveContentBody()) {
+      await waitUntil(haveContentBody, domPollInterval)
     }
 
-    return await action(iframe, iframe.contentWindow)
+    return await action(iframe, iframe.contentWindow as Window)
   } finally {
     iframe.parentNode?.removeChild(iframe)
   }
--- a/src/utils/dom.ts
+++ b/src/utils/dom.ts
@@ -50,11 +50,10 @@ export async function withIframe<T>(
         // The contentWindow.document can miss in JSDOM (https://github.com/jsdom/jsdom).
         if (iframe.contentWindow?.document?.readyState === 'complete') {
           resolve()
-        } else {
-          setTimeout(checkReadyState, 10)
+          clearInterval(timerId)
         }
       }
-      checkReadyState()
+      const timerId = setInterval(checkReadyState, 10)
     })
 
     while (!iframe.contentWindow?.document?.body) {


Finesse:
Hi @lightmare. Thanks for your work. But could you please provide benchmark results or a benchmark itself that demonstrates the difference in performance between the 2 approaches? Why have you concluded that the `setTimeout` functions itself abuses the CPU (instead of `iframe.contentWindow?.document?.readyState` for example)?

`setTimeout` and `setInterval` have several differences to know about while using. But I've never faced and never heard about performance issues of `setTimeout`.

lightmare:
I don't know why exactly this library needs a 100Hz timer, but using `setTimeout` for repeatedly checking a trivial condition at fixed intervals is definitely wrong. Even [MDN setTimeout doc](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#description) suggests using `setInterval` for that purpose.

---

For context, I got the first hint that there might be an issue from Firefox Task Manager (about:performance), which showed "medium" energy impact of Slack tabs all the time. Every other page (I visit) rarely, if ever, goes beyond "low" impact. So I did a performance recording (on app.slack.com/client/..., not their home page), which showed roughly 200ms in a 10-second recording spent in [r gantry-vendors.f8b134c.min.js:70:9164](https://a.slack-edge.com/bv1-9/gantry-vendors.f8b134c.min.js?cacheKey=gantry-1638766647) — that's this function:
```js
var r=function(){var t,n;"complete"===(null===(n=null===(t=c.contentWindow)||void 0===t?void 0:t.document)||void 0===n?void 0:n.readyState)?e():setTimeout(r,10)}
```
It's obviously transpiled optional chaining, so I searched for `.contentWindow?.document?.readyState` on sourcegraph, which lead me here.

---

Now to check fingerprintjs alone, without all the other overhead slack might have, I ran `yarn playground:start --port 1234` and opened it in a new instance of firefox without any plugins. Saved 10-second performance recordings on master and this PR's branch.

[firefox perf recordings](https://ionshare.io/198be1dd-b30a-47e9-a33f-7c42b81dd8ea/download1)

What the Performance tools shows:
- Waterfall tab
  - `profile-master` — tons of `setTimeout` handler calls, each with duration between 0.05ms and 0.30ms, mostly around 0.10ms
  - `profile-use-set-interval` — tons of `setInterval` handler calls, each with duration between 0 and 0.02ms
- Call Tree tab
  - `profile-master` — `checkReadyState` total time: 248ms, self time: 9ms; Tools (`Window.setTimeout`) self time: 38ms;
  - `profile-use-set-interval` — `checkReadyState` total time: 6ms, self time: 4ms;

Now to be fair, in the `profile-master` recording, upon further inspection (enabling Show Gecko Platform Data) I found out that from the 248ms total, most of the time was actually spent in the profiler capturing backtraces (I assume that's because nesting setTimeout is kinda like deep recursion), meaning the whole measurement could be bogus.

Since I had doubts about what the profiler actually measured, I looked again at `about:performance`. It's constantly showing FingerprintJS Playground with "medium" energy impact, and a tooltip saying:
- master branch: "Dispatches in the last seconds: 392 (~20ms)"
- this PR branch: "Dispatches in the last seconds: 392 (~10ms)"

Is that enough proof that nesting `setTimeout` is more costly than a single `setInterval + clearInterval`?

Finesse:
Did you take into account that the playground keeps a websocket connection to the development server? The library itself idles after it prints the result, so it can't affect the performance constantly. If you want a more pure test, please build the playground so that it runs nothing other than the library: `yarn playground:build` and open `index.html` in the browser.

I believe that running `setInterval` can be more power-efficient than running `setTimeout` repeatedly. The library calls `setTimeout` few times, so its portion of the library's energy impact is tiny, and your optimization gives almost no effect.

You use an `as` operator in your code. It decreases the code reliability because it suppresses a TypeScript check. There must be a good reason to decrease the reliability, so I want to see a test that clearly shows that your change does increase the performance of power efficiency noticeably.

A much simpler and more efficient optimization can be done: increase the timer interval. In fact, the interval for the first timer in the `dom.ts` file can be increased because it's a fallback for WeChat that doesn't fire the iframe's `load` event.

-------------

**Update**. I've found out that the repetitive timer in the `dom.ts` file never stops in Firefox because the iframe never goes into the `complete` state. That is a bug that needs to be fixed. The timer must stop when the iframe `load` or `error` event fires. This is the effect of fixing the never-ending loop:

image

I've made a PR that fixes this bug. Thank you for helping finding it.

After this fix the `setTimeout` function is called only 8 times in Firefox. That is a very low number to think about an optimization IMHO. So I think your changes should be rejected. What do you think?

lightmare:
> I've made [a PR](https://github.com/fingerprintjs/fingerprintjs/pull/717) that fixes this bug. Thank you for helping finding it.

Awesome, that fixes it for me. I didn't expect that to be a bug, the high frequency made me think it was something important and intended to work that way.

However I'd still use `setInterval`, because that's a better tool for checking something at fixed intervals. e.g.
```diff
diff --git a/src/utils/dom.ts b/src/utils/dom.ts
index 06b44b7..1d7a8d4 100644
--- a/src/utils/dom.ts
+++ b/src/utils/dom.ts
@@ -26,6 +26,8 @@ export async function withIframe<T>(
   const iframe = d.createElement('iframe')
 
   try {
+    let timerId: undefined | ReturnType<typeof setInterval>
+
     await new Promise((resolve, reject) => {
       iframe.onload = resolve
       iframe.onerror = reject
@@ -50,11 +52,17 @@ export async function withIframe<T>(
         // The contentWindow.document can miss in JSDOM (https://github.com/jsdom/jsdom).
         if (iframe.contentWindow?.document?.readyState === 'complete') {
           resolve()
-        } else {
-          setTimeout(checkReadyState, 10)
         }
       }
-      checkReadyState()
+      timerId = setInterval(checkReadyState, domPollInterval)
+    }).finally(() => {
+      // The ready state never goes to 'complete' in Firefox, so stopping the timer from
+      // within checkReadyState wouldn't be enough. We must clear the interval after the
+      // promise is fulfilled, regardless of how that happened.
+      // See https://github.com/fingerprintjs/fingerprintjs/pull/716#issuecomment-986898796
+      if (timerId) {
+        clearInterval(timerId)
+      }
     })
 
     while (!iframe.contentWindow?.document?.body) {
```


> You use an `as` operator in your code. It decreases the code reliability because it suppresses a TypeScript check.

Agreed. I should include a comment, and better use suffix `!` instead of typecast. It was needed because previously the while loop acted as a non-null guard on `iframe.contentWindow`, but when I wrapped the condition in a function, TS wasn't able to infer that it cannot be null.


> After this fix the `setTimeout` function is called only 8 times in Firefox. That is a very low number to think about an optimization IMHO. So I think your changes should be rejected. What do you think?

I'm fine with this PR getting rejected. But I maintain that the pattern `while (!x) await wait(t)` is just wrong. It doesn't matter whether it calls `setTimeout` 8 or 800 times. Every one of those calls is a misuse, and all but the first are overhead (however minor, it's a straw the camel doesn't have to carry).

Finesse:
@lightmare The PR was closed automatically. Feel free to reopen the PR.

> However I'd still use setInterval, because that's a better tool for checking something at fixed intervals. e.g.

The change that you show in your comment looks good. But there is a couple of notices:

- The first check should be done synchronously (`setInterval` doesn't do it). The iframe is ready immediately in WeChat, so waiting for 50ms reduces the performance.
- The project's code sticks to the `async/await` syntax, so please use `try {} finally {}` instead or move interval cleanup inside the promise constructor (like I did in [my PR](https://github.com/fingerprintjs/fingerprintjs/pull/717))
- The `timerId` value may be `0`. Even though the documentation says the returned value is a positive integer, there can be browsers with a non-standard implementation. So the check for clearing the interval should be `timerId !== undefined` instead of `timerId` just in case.

> Agreed. I should include a comment, and better use suffix `!` instead of typecast. It was needed because previously the while loop acted as a non-null guard on `iframe.contentWindow`, but when I wrapped the condition in a function, TS wasn't able to infer that it cannot be null.

The `!` operator is not a good solution too because it suppress a TypeScript check. The current `while () await` syntax was chosen intentionally to act as a type guard. Consider using the `is` operator to make a function that acts as a type guard. If it is not possible, or the code becomes much more complex than it is now, then it won't be worth replacing `setTimeout` with `setInterval`.

> But I maintain that the pattern `while (!x) await wait(t)` is just wrong. It doesn't matter whether it calls `setTimeout` 8 or 800 times. Every one of those calls is a misuse, and all but the first are overhead (however minor, it's a straw the camel doesn't have to carry).

I understand. But this is not a case to worry about, because in contrast to usual `setInterval` use cases, the `while (!x) await wait(t)` loops in the code are expected to run few iterations (0 or 1 in most cases). Optimizing this is not worth sacrificing the code reliability, simplicity and size.

PR #722, opened 2021-12-21
Reduce the number of NPM monitoring requests

Finesse:
Send the requests on every 100 runs instead of every run


bayotop:
Looks good. Was just wondering if we should add a mention to the [API docs](https://github.com/fingerprintjs/fingerprintjs/blob/master/docs/api.md) about this behaviour?

bayotop (approved):

Finesse:
> Send the requests on every 100 runs instead of every run

I wondered if it makes sense. But because you've mentioned it, I've decided to add it.

PR #729, opened 2022-01-08
Added airbnb eslint rules and a video card entropy source

spectercs:
Hello,
I want to apologize in advance as I don't really know the contribution guidelines for this repository.
And I tried to follow contributing.md as much as I could.
Thank you for this very helpful package !

Things I have added

- Video Card entropy source

Things I have modified

- Changed early use functions to conform with this rule https://eslint.org/docs/rules/no-use-before-define
- Removed restricted globals like (screen) to (window.screen) https://eslint.org/docs/rules/no-restricted-globals
- Replaced for..in with .forEach https://eslint.org/docs/rules/no-restricted-syntax

DevPackages added:

- eslint-config-airbnb
- eslint-config-airbnb-typescript
- eslint v8.6.0


Finesse:
Hi @spectercs. I appreciate the effort that you've put into this PR and thank you, but we can't merge it.

First of all, individual features, enhancements and fixes should be published in individual PRs. Separate PRs are easier to review and discuss, and can be merged independently. In this case you should make 2 PRs: one with the new entropy source and one with the ESLint changes.

Second, the way the code is checked is subjective. So we don't accept code style and linter configuration changes unless there is a clear benefit. For each of the linter change, please proof that the change is necessary or remove it.

The new entropy source looks interesting. I have a few notices:

- Have you tested the entropy source in various browsers and devices? Does the result change over browsers and devices, can you provide an example? Is the source stable, i.e. does it always give the same result on the same browser and device?
- Why have you decided to use only one WebGL parameter? The StackOverflow answer that you've referenced suggests using `UNMASKED_VENDOR_WEBGL` too. Why not to use the other [WebGL parameters](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/getParameter)? As I mentioned above, there should be only parameters that give stable and unique result.
- Please make a unit test for the entropy source. You can find many test examples in the [sources](https://github.com/fingerprintjs/fingerprintjs/tree/master/src/sources) directory. The test will run in various browsers using BrowserStack.

spectercs:
Thank you for your kind response. I'll open a new pull request with only the necessary modifications.

To answer your questions:

- Upon testing it gives stable results across browsers and devices it gives stable results
HOWEVER chromium based browsers give really precise result on the device graphics card, Brave browser even returns the dedicated graphics card name (funny as it is a privacy focused browser). firefox returns an older version of the card. (i'll give a detailed example in my next pull request).
Yet the results are stable !

- I ignored UNMASKED_VENDOR_WEBGL as it only returns the name of the vendor with the name of the browser (ex: (Google inc NVIDIA) or Google Inc Firefox. I'll re add it in the next pull request !
- Upon further inspections the withIframe is not needed for this entropy source

PR #737, opened 2022-01-28
fix: The library triggers a console error message unless the CSP allows `unsafe-inline` for `style-src`

### LINKED ISSUE #733: fingerprintjs violates the CSP header while trying to set inline-style
### Scenario

- What's the scenario, what happens and what did you expect to happen? 
--> Loading page which uses fingerprintjs but doesn't allow `inline styles`
- What device and browser are you using? 
--> Google Chrome on a Mac
- What version of fingerprintjs are you using?
--> 3.3.2

### Fingerprint data

We have very strict policies on our page and are not allowing `inline-styles`. 
But while using fingerprintjs, the library tries to set some attributes and violates our CSP header. 
Currently i don't understand why fingerprintjs is trying to change something without my knowledge. 

Here is the function of fingerprintjs which causes trouble.
```
function R(e) {
                for (var t = function(e) {
                    for (var t, n, r = "Unexpected syntax '" + e + "'", i = /^\s*([a-z-]*)(.*)$/i.exec(e), o = i[1] || void 0, a = {}, s = /([.:#][\w-]+|\[.+?\])/gi, c = function(e, t) {
                        a[e] = a[e] || [],
                        a[e].push(t)
                    }; ; ) {
                        var u = s.exec(i[2]);
                        if (!u)
                            break;
                        var l = u[0];
                        switch (l[0]) {
                        case ".":
                            c("class", l.slice(1));
                            break;
                        case "#":
                            c("id", l.slice(1));
                            break;
                        case "[":
                            var p = /^\[([\w-]+)([~|^$*]?=("(.*?)"|([\w-]+)))?(\s+[is])?\]$/.exec(l);
                            if (!p)
                                throw new Error(r);
                            c(p[1], null !== (n = null !== (t = p[4]) && void 0 !== t ? t : p[5]) && void 0 !== n ? n : "");
                            break;
                        default:
                            throw new Error(r)
                        }
                    }
                    return [o, a]
                }(e), n = t[0], r = t[1], i = document.createElement(null != n ? n : "div"), o = 0, a = Object.keys(r); o < a.length; o++) {
                    var s = a[o];
                    i.setAttribute(s, r[s].join(" "))
                }
                return i
            }
```

The last line `i.setAttribute(s, r[s].join(" "))` is the issue.
Does someone know why this function is called and what it does and is the only solution to allow `inline style` while using fingerprintjs?

Finesse:
Resolves #733

A keyword for future searches: Content Security Policy

### DIFF
--- a/src/utils/dom.ts
+++ b/src/utils/dom.ts
@@ -96,10 +96,25 @@ export function selectorToElement(selector: string): HTMLElement {
     // Changing the `style` attribute can cause a CSP error, therefore we change the `style.cssText` property.
     // https://github.com/fingerprintjs/fingerprintjs/issues/733
     if (name === 'style') {
-      element.style.cssText = value
+      addStyleString(element.style, value)
     } else {
       element.setAttribute(name, value)
     }
   }
   return element
+}
+
+/**
+ * Adds CSS styles from a string in such a way that doesn't trigger a CSP warning (unsafe-inline or unsafe-eval)
+ */
+export function addStyleString(style: CSSStyleDeclaration, source: string): void {
+  // We don't use `style.cssText` because browsers must block it when no `unsafe-eval` CSP is presented: https://csplite.com/csp145/#w3c_note
+  // Even though the browsers ignore this standard, we don't use `cssText` just in case.
+  for (const property of source.split(';')) {
+    const match = /^\s*([\w-]+)\s*:\s*(.+?)(\s*!([\w-]+))?\s*$/.exec(property)
+    if (match) {
+      const [, name, value, , priority] = match
+      style.setProperty(name, value, priority)
+    }
+  }
 }
--- a/src/utils/dom.ts
+++ b/src/utils/dom.ts
@@ -92,7 +92,14 @@ export function selectorToElement(selector: string): HTMLElement {
   const [tag, attributes] = parseSimpleCssSelector(selector)
   const element = document.createElement(tag ?? 'div')
   for (const name of Object.keys(attributes)) {
-    element.setAttribute(name, attributes[name].join(' '))
+    const value = attributes[name].join(' ')
+    // Changing the `style` attribute can cause a CSP error, therefore we change the `style.cssText` property.
+    // https://github.com/fingerprintjs/fingerprintjs/issues/733
+    if (name === 'style') {
+      element.style.cssText = value
+    } else {
+      element.setAttribute(name, value)
+    }
   }
   return element
 }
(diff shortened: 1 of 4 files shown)


src/utils/dom.ts
```diff
@@ -92,7 +92,14 @@ export function selectorToElement(selector: string): HTMLElement {
   const [tag, attributes] = parseSimpleCssSelector(selector)
   const element = document.createElement(tag ?? 'div')
   for (const name of Object.keys(attributes)) {
-    element.setAttribute(name, attributes[name].join(' '))
+    const value = attributes[name].join(' ')
+    // Changing the `style` attribute can cause a CSP error, therefore we change the `style.cssText` property.
+    // https://github.com/fingerprintjs/fingerprintjs/issues/733
+    if (name === 'style') {
+      element.style.cssText = value
```
bayotop, line 99: I think the safest option here (regarding CSP) would be to parse the value and then use `.style.<specific-property> = 'single value'`. The reason is that `.cssText` requires the browser to parse the provided string and [according to the spec this should be gated by `unsafe-eval`](https://www.w3.org/TR/CSP3/#directive-style-src). Also MDN lists this operation as [unsafe](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src#unsafe_style_expressions).

Regardless, I tested it in latest Firefox and Chrome, and `HTMLElement.style.cssText` doesn't seem to be blocked if `unsafe-eval` is missing, which might be explained by [this w3c comment](https://csplite.com/csp145/#w3c_note):

> HTMLElement.style (eg el.style.color='green';) should not be blocked in the absence of 'unsafe-eval' nor in the absence of 'unsafe-inline'

but: 

> .cssText, .insertRule() are meant to block in the absence of 'unsafe-eval'.

So I think we have two options:
1. Keep using `.cssText` and get back to this if / when browsers change their behaviour to match the specification (if I'm reading it correctly).
2. Parse the selectors and use the specific style property with a single value (this might be too difficult to do reliably, but covering only the few `style` occurrences that we currently have in the rules shouldn't be too hard I think).

wdyt?

Finesse: Thank you for the findings. I've also found that the Chromium team [tried to implement](https://bugs.chromium.org/p/chromium/issues/detail?id=336413) the `unsafe-eval` restriction, but they rolled back the changes because it broke many websites.

Even though the `unsafe-eval` policy enablement is very unlikely, I've implemented a solution that doesn't use `cssText` to reduce the risk.

bayotop (approved):

PR #738, opened 2022-02-08
Enums added

yaritaft:
Added Enums to remove Hardcoded OS, Browsers, Engines, Engine Types


yaritaft:
> Hi @yaritaft, thank you for the effort. The code style is subjective, so we don't accept code style changes unless there is a clear benefit. Please demonstrate that the change is necessary.
> 
> The only benefit I see is deduplicating the code in the `browserstackBrowsers` declaration. But Karma won't be able to import the TS file with the enums from the JS configuration file, so it's more convenient to replace the enum file with constants declared right in the `karma.browserstack.config.js` file. If you agree, please make constants for the other duplications in `browserstackBrowsers` where it makes sense.

The idea was to avoid copied and pasted values and move them to enums or constants in order to reuse them in other places without having to copy and paste everywhere.

Finesse:
@yaritaft 

> The idea was to avoid copied and pasted values and move them to enums or constants in order to reuse them in other places without having to copy and paste everywhere.

In software development we don't create enums just because some values repeat. We create constants and enums for **semantically equal** values. That is, 2 values may be the same but mean different things. Even if they both mean an OS name, they are semantically different because they are addressed to different libraries and these libraries may accept different names.

Please do these changes:

1. Revert the changes in `tests/utils/index.ts` and `tests/utils/enums.ts` because there is no duplication to remove
2. There is no need to `Object.freeze` in `tests/karma.browserstack.config.js`

Finesse:
Closing due to the lack of response

PR #774, opened 2022-06-22
Feature | Detect if MetaMask extension is installed

kalexey89:
[MetaMask](https://chrome.google.com/webstore/detail/metamask/nkbihfbeogaeaoehlefnkodbefgpgknn) is a popular crypto wallet that works as a browser extension. We can detect that a given extension is installed and use this knowledge as an entropy source.


kalexey89:
Yes indeed. Understood, thank You @Finesse.

Issue #781, opened 2022-07-04
Ads blocker detects it and blocks from accessing the module 


Arjun-Ghimire:
### Ads blocker detects it and blocks from accessing the module 

![image](https://user-images.githubusercontent.com/82378187/177088531-4f3c7f19-b553-4ee9-9407-86e27f2c931e.png)

### Fingerprint data

Adblocker is preventing from loading loading the data from the domain openfpcdn.io please do something to solve this! thanks!

Arjun-Ghimire:
It doesn't work with ublock origin ads blocker please fix this asap!
https://chrome.google.com/webstore/detail/ublock-origin/cjpalhdlnbpafiamejdnhcphjbkeiagm?hl=en

Valve:
@Arjun-Ghimire thanks for letting us know about this issue. We cannot make changes to an external product such as uBlock.
You have several options to fix this:
-- [Use the NPM installation option for FingerprintJS](https://github.com/fingerprintjs/fingerprintjs/blob/master/docs/api.md#webpackrollupnpmyarn)
-- [Use the free tier for FingerprintJS Pro](https://dashboard.fingerprint.com/signup), which allows to make up to 20K identification events per month. This option supports both the CDN and NPM installation options as well as [custom subdomain](https://dev.fingerprint.com/docs/subdomain-integration) that will prevent the uBlock (or any ad blocker) blocking
-- [Use the paid tier for FingerprintJS Pro](https://dashboard.fingerprint.com/signup), which allows to make unlimited number of identification events per month. This option supports both CDN and NPM installation options and the [custom subdomain](https://dev.fingerprint.com/docs/subdomain-integration) functionality too.

Arjun-Ghimire:
Hello valve,
I am using asp.net core blazor. There is no option for npm install and I don't want to use FingerpintJS pro so can you please help me with the process how I can fix this problem please?

Valve:
@Arjun-Ghimire if you cannot use an NPM package, the only option is to self-host the script on your own domain. Note that it will make your script quickly outdated and it will stop working with new browsers as they get released in the future.

Arjun-Ghimire:
Hello Valve, I already tried by self hosting the script but still it doesn't works which means there is some content on the script itself that is preventing it from ads blocker may be you need to review? 

I just need to point this     

const fpPromise = import('https://openfpcdn.io/fingerprintjs/v3')
        .then(FingerprintJS => FingerprintJS.load());

[...]

PR #782, opened 2022-07-05
Create bypass_adblock_detection.md

makma:
## Motivation
- https://github.com/fingerprintjs/fingerprintjs/issues/781
- https://stackoverflow.com/questions/72859456/ublock-origin-prevents-my-js-script-to-load

### DIFF
--- a/docs/bypass_adblock_detection.md
+++ b/docs/bypass_adblock_detection.md
@@ -1,0 +1,16 @@
+# Bypass ad blockers detection
+
+Some ad blockers might detect and block this library because of the known CDN URL or the script name.
+To bypass the specific detection, one needs to check the applied rules.
+Here are some generic options to circumvent this behavior.
+
+## Open-source alternatives
+- Use the [NPM installation option](https://github.com/fingerprintjs/fingerprintjs/blob/master/docs/api.md#webpackrollupnpmyarn) for the open source FingerprintJS. Additionally, make sure the script is not served from the `*/fingerprint.js` path.
+- The less preferred way is to self-host the script on your own domain. It will make your script quickly outdated and it will stop working with new browsers as they get released in the future.
+You can find scripts on the following URLs (replace the version with the current version e.g. with `v3.3.3`).
+  - ES Module download URL: `https://openfpcdn.io/fingerprintjs/v3.3.3/esm.min.js`.
+  - Universal Module Definition download URL: `https://openfpcdn.io/fingerprintjs/v3.3.3/umd.min.js`.
+
+## Commercial alternatives
+- Use the [free tier for FingerprintJS Pro](https://dashboard.fingerprint.com/signup), which allows making up to 20K identification events per month. This option supports both the CDN and NPM installation options as well as a [custom subdomain](https://dev.fingerprint.com/docs/subdomain-integration) that will prevent the ad blocker blocking.
+- Use the [paid tier for FingerprintJS Pro](https://dashboard.fingerprint.com/signup), which allows making an unlimited number of identification events per month. This option supports both CDN and NPM installation options and the [custom subdomain](https://dev.fingerprint.com/docs/subdomain-integration) functionality too.


docs/bypass_adblock_detection.md
```diff
@@ -0,0 +1,16 @@
...
+Some ad blockers might detect and block this library because of the known CDN URL or the script name.
+To bypass the specific detection, one needs to check the applied rules.
+Here are some generic options to circumvent this behavior.
+
+## Open-source alternatives
+- Use the [NPM installation option](https://github.com/fingerprintjs/fingerprintjs/blob/master/docs/api.md#webpackrollupnpmyarn) for the open source FingerprintJS. Additionally, make sure the script is not served from the `*/fingerprint.js` path.
+- The less preferred way is to self-host the script on your own domain. It will make your script quickly outdated and it will stop working with new browsers as they get released in the future.
+You can find scripts on the following URLs (replace the version with the current version e.g. with `v3.3.3`).
+  - ES Module download URL: `https://openfpcdn.io/fingerprintjs/v3.3.3/esm.min.js`.
+  - Universal Module Definition download URL: `https://openfpcdn.io/fingerprintjs/v3.3.3/umd.min.js`.
```
Finesse, line 12: Version-unspecific URLs should be used:

```suggestion
  - ES Module download URL: `https://openfpcdn.io/fingerprintjs/v3/esm.min.js`.
  - Universal Module Definition download URL: `https://openfpcdn.io/fingerprintjs/v3/umd.min.js`.
```

Finesse: "UMD" is known better than "Universal Module Definition". Maybe it will be unclear for unexperienced developers which version to download.

Finesse (changes requested):

docs/bypass_adblock_detection.md
```diff
@@ -0,0 +1,16 @@
+# Bypass ad blockers detection
```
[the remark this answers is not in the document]
makma: Exactly, sounds good to me.

docs/bypass_adblock_detection.md
makma: Fixed in https://github.com/fingerprintjs/fingerprintjs/pull/782/files#diff-ca75138eebfa7215a976ae8804b36955e154c883deafffc140fd52580b8631dc

docs/bypass_adblock_detection.md
```diff
@@ -0,0 +1,16 @@
...
+
+Some ad blockers might detect and block this library because of the known CDN URL or the script name.
+To bypass the specific detection, one needs to check the applied rules.
+Here are some generic options to circumvent this behavior.
+
+## Open-source alternatives
+- Use the [NPM installation option](https://github.com/fingerprintjs/fingerprintjs/blob/master/docs/api.md#webpackrollupnpmyarn) for the open source FingerprintJS. Additionally, make sure the script is not served from the `*/fingerprint.js` path.
```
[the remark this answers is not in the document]
makma: Fixed in https://github.com/fingerprintjs/fingerprintjs/pull/782/commits/dbe4b7ff48bd097839f2709726a4c10ea9a18723

docs/bypass_adblock_detection.md
Finesse: Looks good now. Could you please also add a link to this article to the API guide or to the readme?

makma: - After a tiny internal survey, I decided to change `bypass` to `evade` wording. I renamed all occurrences.
- I added ad blockers evasion to the feature matrix. The most natural place to me was in the `Additional Features` section right before the `Geolocation`.
- Re: adding a link. I didn't want to add a new unnecessary section to the readme. Also, adding to the API guide is too deep IMO. I decided to add a link to the feature matrix. Once somebody looks for `ad blockers` they immediately see the link.

readme.md
```diff
@@ -93,6 +93,7 @@ Full product comparison:
     <tr><td><b>Incognito mode detection</b><br/><i>works in all modern browsers - see our full list of <a href="https://dev.fingerprint.com/docs/browser-support/" target="_blank">browsers supported</a></i></td><td align="center">–</td><td align="center">✓</td></tr>
     <tr><td><b>Server-side accuracy increase</b><br/><i>based on additional server-side signals, such as TLS crypto support, ipv4/v6 data and others</i></td><td align="center">–</td><td align="center">✓</td></tr>
     <tr><td><b>Query API & realtime Webhooks</b><br/><i>build flexible workflows</i></td><td align="center">–</td><td align="center">✓</td></tr>
+    <tr><td><b>Ability to evade ad blockers</b></td><td align="center"><a href="docs/evade_ad_blockers.md" target="_blank">manual</a></td><td align="center">automatic (via custom domains)</td></tr>
```
Finesse, line 96: I don't think the comparison table is a good place to put this link because Pro version requires more effort to evade ad blockers

Finesse (changes requested):

readme.md
Finesse: In fact, the comparison table is a good place, but I don't agree that evasion in Pro version is automatic

makma: What do you suggest? Something like `out of the box (subdomain setup)`, `supported (subdomain setup)`,  or just `via subdomain setup`?

Finesse: Maybe?

| Open | Pro |
|--------|--------|
| supported (NPM, own hosting) | supported (NPM + subdomain) |

Both cells can include links BTW

makma: I don't think it's fair to say `supported` for both. There are significant pros for Pro and the `supported` label for both of them might imply a false feeling of equality (which is not true IMO). I mean, for open-source, this `supported` term means you need to implement it by yourself, and additionally, you have to give up automatic updates of the fingerprinting script because you need to use the static NPM package. On the contrary, this doesn't apply to Pro.

At first, I wanted to add also a link for the subdomain setup, nevertheless, I decided not to since we don't link Pro's features from the feature matrix at all.

makma: Something similar is proposed in the PR.

makma: Yes, you are right @Finesse with both (the NPM approach and Cloudflare integration capabilities). I was a little bit confused with the loader capabilities and its implications.

Finesse: Both "manual + automatic" and "manual + manual" are ok for me, because the table is not required to be 100% accurate. Please merge the PR with the variant that you like.

PR #793, opened 2022-08-04
Fix "use identification for fraud" in the readme

### LINKED ISSUE #792: No, don't use it for fraud
From the README:

> Fingerprint Pro is GDPR and CCPA compliant as the data processor. You still need to be compliant as the data controller and *use the identification for fraud* under legitimate interest or ask for user consent.

I don't think that's what it's supposed to say. It looks like it should say "use the identification for fraud prevention".

Finesse:
Resolves #792


makma (changes requested):
Needs to be fixed also here: https://github.com/fingerprintjs/fingerprint.com/blob/0130b8a0466a393a6ef10f33d473356bf3192476/src/components/githubPage/FeatureComparisonSection/FeatureComparisonSection.tsx#L116

But that's a different repo, sorry.

makma (approved):

PR #799, opened 2022-08-17
build[package.json]: move tslib to devDependencies

Valar103769:
esbuild `s outfile can include  tslib,  it is not necessary

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,9 +43,6 @@
     "check:dts": "tsc --isolatedModules --noEmit dist/fp.d.ts",
     "check:ssr": "node --require './dist/fp.cjs.js' --eval '' || (echo \"The distributive files can't be used with server side rendering. Make sure the code doesn't use browser API until an exported function is called.\" && exit 1)"
   },
-  "dependencies": {
-    "tslib": "^2.0.1"
-  },
   "devDependencies": {
     "@rollup/plugin-json": "^4.1.0",
     "@rollup/plugin-node-resolve": "^13.3.0",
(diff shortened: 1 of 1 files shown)


Finesse:
Hello @Valar103769. Thank you for contributing. What problem do you try to solve with this PR? How is esbuild related to FingerprintJS?

I can't merge this PR because it breaks the library. [The JS file that we provide for the NPM installation method](https://cdn.jsdelivr.net/npm/@fingerprintjs/fingerprintjs@3.3.4/dist/fp.esm.js) imports `tslib`, so if you remove the `tslib` dependency from `package.json`, the end users won't install `tslib` (unless it's required by another dependency) and will get a "missing dependency" error during building their code.

Valar103769:
> Hello @Valar103769. Thank you for contributing. What problem do you try to solve with this PR? How is esbuild related to FingerprintJS?
> 
> I can't merge this PR because it breaks the library. [The JS file that we provide for the NPM installation method](https://cdn.jsdelivr.net/npm/@fingerprintjs/fingerprintjs@3.3.4/dist/fp.esm.js) imports `tslib`, so if you remove the `tslib` dependency from the `dependencies` field of the `package.json`, the end users won't install `tslib` (unless it's required by another dependency) and will get a "missing dependency" error during building their code.

i think tslib is a devdependencies instead of  dependncies

Finesse:
> i think tslib is a devdependencies instead of dependncies

@Valar103769 It depends. `dependencies` is for packages that are necessary for the library users to run the library. FingerprintJS requires `tslib` because [its JS file](https://cdn.jsdelivr.net/npm/@fingerprintjs/fingerprintjs@3.3.4/dist/fp.esm.js) imports `tslib` (see the line below the comment). If somebody installs FingerprintJS and doesn't install `tslib`, the library will fail to run because it will try to import a missing `tslib` dependency.

For this reason I'm closing the PR, sorry.

PR #802, opened 2022-08-20
Run terser twice

jeetiss:
Run terser twice for better compression it's not a big deal but every bit matters

### size

```js
// file       before   after
fp.min.js      32183 → 31892
fp.umd.min.js  32374 → 32083
```

### DIFF
--- a/terser.config.ts
+++ b/terser.config.ts
@@ -1,6 +1,9 @@
 // Note: this file is non standard, it's imported explicitly
 // Reference: https://github.com/terser/terser
 export default {
+  compress: {
+    passes: 2,
+  },
   format: {
     comments: false,
   },


Finesse:
Hello @jeetiss. Thank you for putting your effort into such a small detail. But I'm afraid it won't be appreciated, here is why. There are 2 recommended installation methods for FingerprintJS:

The first is our CDN (`import('https://openfpcdn.io/fingerprintjs/v3')`). The code received by this endpoint is built by the CDN in runtime. As you can see in [the CDN source code](https://github.com/fingerprintjs/cdn), the CDN uses the JS file pointed by the `module` field of [package.json](https://github.com/fingerprintjs/fingerprintjs/blob/master/package.json). This JS file is `dist/fp.esm.js` which is not minified. Therefore your change won't affect the CDN.

The CDN minifies the JS code in runtime. You may want to do a similar change in the CDN, but I can say in advance that we won't accept such change because the minification takes significant time, and performance is very important to the CDN.

The second installation method is NPM or Yarn (`npm i @fingerprintjs/fingerprintjs`). The `dist/fp.esm.js` file will be used for the same reason as described above, so your changes won't affect the result too. The minification in this case is performed by the users of FingerprintJS.

The only reason for FingerprintJS `build` command to produce the minified files is to keep the old jsDelivr installation method (https://cdn.jsdelivr.net/npm/@fingerprintjs/fingerprintjs@3/dist/fp.min.js) working.

Nevertheless, I see no big reason not to merge the PR. What do you think?

jeetiss:
@Finesse, sure! Thanks for that detailed explanation

PR #852, opened 2022-12-21
Amend the readme

Finesse:
- Remove an unused image
- Split the description from the accuracy note. I think it's important to have a simple answer to a "why may I need this" question. A single short paragraph is used for this.
- Move a link to a React usage example from the Pro section to an OSS section. Also change the icon to an emoji resembling React logo.
- Stick to the common file naming style
- Add timestamps to the Pro response example
- Hide a Pro response example under a spoiler, because it's bulky, and an example is already given on the Pro picture
- Change "FingerprintJS Pro" to "Fingerprint Pro"


Valve (approved):

Finesse:
@makma Do you want to take a look before I merge?

makma:
Thank you @Finesse, looks good to me.

PR #864, opened 2023-01-18
feature: Add architecture source based on NaN output

bayotop:


src/sources/architecture.test.ts
```diff
@@ -0,0 +1,10 @@
+import getArchitecture from './architecture'
+
+describe('Sources', () => {
+  describe('getArchitecture', () => {
```
Finesse, line 4: Please use `'architecture'` like in the other tests

Finesse (changes requested):

Finesse:
Asking just in case: have you checked that the new signal is really stable?

bayotop:
> Asking just in case: have you checked that the new signal is really stable?

I tested on MacOS 12.3 and iOS 16.2 in Safari, Windows 10 and Android 12 across Chrome, Firefox and Edge in the latest versions in private and normal mode windows. Also, same value for both `window.isSecureContext` options. The results were all consistent as expected.

I can't think of anything that would make the returned value unstable.

Finesse (approved):

PR #875, opened 2023-02-20
chore: Separate compiler configuration for ts-node

dj-stormtrooper:


playground/webpack.config.ts
```diff
@@ -17,7 +17,7 @@ const configurationFactory: ConfigOptions = (_env, { mode = 'development' }) =>
         loader: 'ts-loader',
         exclude: /node_modules/,
         options: {
-          configFile: '../tsconfig.browser.json',
+          configFile: '../tsconfig.json',
```
Finesse, line 20: This option is probably unnecessary because TypeScript [should search](https://github.com/TypeStrong/ts-loader#configfile) for `tsconfig.json` in the parent directories automatically:

> The loader then will search for the config file of each entry point in the respective entry point's containing folder. If a config file cannot be found there, it will travel up the parent directory chain and look for the config file in those folders.

tsconfig.json
```diff
@@ -1,9 +1,9 @@
 // Common style settings and default options for ts-node use-cases
```
Finesse, line 1: This comment needs an actualization

Finesse (changes requested):

playground/webpack.config.ts
dj-stormtrooper: Ok, thanks, I'll check

Finesse (approved):

PR #887, opened 2023-04-03
chore: Moves retries configuration to separate file

dj-stormtrooper:


fpkamp:
@Finesse maybe we should do the same in our example project?

Finesse:
> @Finesse maybe we should do the same in our example project?

@fpkamp This is not important. But please do it if you don't mind.

Finesse (approved):

PR #895, opened 2023-04-21
use latest versions of safari for OSX tests

fpkamp:


karma.conf.ts
```diff
@@ -49,9 +49,9 @@ const browserstackBrowsers = {
...
   // Windows11_FirefoxLatest_Incognito: { platform: 'Windows', osVersion: '11', browserName: 'Firefox', browserVersion: 'latest-beta, ...firefoxIncognitoCapabilities },
   Windows11_EdgeLatest: { platform: 'Windows', osVersion: '11', browserName: 'Edge', browserVersion: 'latest-beta', useHttps: true },
-  'OSX10.14_Safari12': { platform: 'OS X', osVersion: 'Mojave', browserName: 'Safari', browserVersion: '12', useHttps: true },
-  OSX12_Safari15: { platform: 'OS X', osVersion: 'Monterey', browserName: 'Safari', browserVersion: '15', useHttps: false },
-  OSX13_Safari16: { platform: 'OS X', osVersion: 'Ventura', browserName: 'Safari', browserVersion: '16', useHttps: false },
+  'OSX10.14_Safari12': { platform: 'OS X', osVersion: 'Mojave', browserName: 'Safari', browserVersion: 'latest', useHttps: true },
+  OSX12_Safari15: { platform: 'OS X', osVersion: 'Monterey', browserName: 'Safari', browserVersion: 'latest', useHttps: false },
+  OSX13_Safari16: { platform: 'OS X', osVersion: 'Ventura', browserName: 'Safari', browserVersion: 'latest', useHttps: false },
```
Finesse, line 54: We want to target specific Safari versions, not macOS versions. Safari major version isn't bound to a macOS version (for example, I have Safari 16.4 on macOS 11 which was shipped with Safari 14 originally). It doesn't seem to be a case on BrowserStack, but still this is a concern to me.

What do you think if we use specific major versions for old versions of macOS, and `latest` for the latest version of macOS?

As I can see in [BrowserStack configurator](https://www.browserstack.com/docs/automate/javascript-testing/configure-test-run-options), there is only 1 Safari version for each macOS version. Did you check that using `latest` instead of `16` makes a difference?

fpkamp: no, it doesn't and you always get served the newest one available for the requested OS X version (for Monterey that would be 15.6 as 16 is not available for Monterey), so I think we don't need it

Finesse: > no, it doesn't

Do you mean that using `latest` instead of a number makes no difference? If so, let's close the PRs.

PR #897, opened 2023-04-28
Automatically retry unit tests if they fail

fpkamp:
So far, the only failures we have experienced were fixed by retrying manually, so in order to reduce the reason to react to false failures, we could use a retry. 
More on this action here:
https://github.com/marketplace/actions/retry-step


Finesse (approved):
LGTM

Please do the same change in the pro repository.

PR #901, opened 2023-05-05
tests: Add unit tests for expected values

bayotop:


src/sources/audio.test.ts
```diff
@@ -6,7 +6,7 @@ describe('Sources', () => {
     it('returns expected value type depending on the browser', async () => {
       const result = getAudioFingerprint()
 
-      if (isSafari() && isMobile() && (getBrowserMajorVersion() ?? 0) < 12) {
+      if (isUnsupportedBrowser()) {
         // WebKit has stopped telling its real version in the user-agent string since version 605.1.15,
         // therefore the browser version has to be checked instead of the engine version.
```
Finesse, line 11: This comment should be inside the `isUnsupportedBrowser` function

src/sources/audio.test.ts
```diff
@@ -21,5 +21,25 @@ describe('Sources', () => {
...
+      if (isUnsupportedBrowser()) {
+        return
+      } else {
+        if (typeof result !== 'function') {
+          throw new Error('Expected to be a function')
+        }
+
+        const first = await result()
+        const second = await result()
+        expect(second).toBe(first)
```
Finesse, line 37: All calls of the same `result` function return literally the same promise, so this assertion can't fail. Therefore `getAudioFingerprint` should be called twice instead.

Finesse (changes requested):
Good job, thank you. Just a couple of minor issues.

src/sources/audio.test.ts
bayotop: Good catch, updated. Thanks!

Finesse (approved):

Finesse (approved):

PR #904, opened 2023-05-13
feature: Add Media Recorder Source

karavaan:
The `MediaRecorder.isTypeSupported` method determines which MIME types are supported by a user's browser to be recorded to a blob.

### DIFF
--- a/src/sources/index.ts
+++ b/src/sources/index.ts
@@ -34,6 +34,7 @@ import getFontPreferences from './font_preferences'
 import getVideoCard from './video_card'
 import isPdfViewerEnabled from './pdf_viewer_enabled'
 import getArchitecture from './architecture'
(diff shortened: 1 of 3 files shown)


karavaan:
@Finesse could you take a look at the pr's I opened?

Finesse:
Hey @karavaan. Thank you for the contribution. I'm sorry for a late response, I was very busy.

I need to make sure the entropy source is good enough before I can include it into the library:

- Does the returned value depend on anything other than the browser and its version (device hardware, browser settings, OS settings, etc)? If yes, how can I reproduce it?
- Does the returned value stay the same in incognito/private mode?

Finesse:
Closing because of lack of response. Feel free to reopen.

PR #905, opened 2023-05-13
feature: Add supported media constraints source

karavaan:
The getSupportedConstraints() method of the [MediaDevices](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices) interface returns an object based on the [MediaTrackSupportedConstraints](https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackSupportedConstraints) dictionary, whose member fields each specify one of the constrainable properties the [user agent](https://developer.mozilla.org/en-US/docs/Glossary/User_agent) understands.


Finesse:
Closing because of lack of response. Feel free to reopen.

PR #908, opened 2023-06-05
chore: Simplify local tests setup

dj-stormtrooper:


Finesse (approved):
Please do this change in https://github.com/fingerprintjs/BotD too. They don't use test retries, there is no need to add it.

dj-stormtrooper:
> Please do this change in https://github.com/fingerprintjs/BotD too. They don't use test retries, there is no need to add it.

@Finesse they use it here
https://github.com/fingerprintjs/BotD/blob/main/tests/integration.test.ts#L9

PR #909, opened 2023-06-06
chore: Simplify local tests setup

dj-stormtrooper:
Previous `tests/utils/karma_global_setup_retries.ts` file was included to bundle unconditionally by more common globs, I moved it to separate directory


karma.conf.ts
```diff
@@ -72,7 +72,7 @@ function makeBuildNumber() {
 }
 
 function setupLocal(config: Config) {
-  const ciSpecificFiles = ['tests/utils/karma_global_setup_retries.ts']
+  const ciSpecificFiles = ['config/karma/karma_global_setup_retries.ts']
```
Finesse, line 75: Creating a root-level directory for a single file is too much. Let's use `resources/karma` instead?

Finesse (changes requested):

karma.conf.ts
dj-stormtrooper: I was thinking we could use this directory for similar configs

Finesse (approved):

PR #912, opened 2023-06-22
use detailed karma logging only when CI has debugging set

fpkamp:


.github/workflows/test.yml
```diff
@@ -41,4 +41,4 @@ jobs:
         env:
           BROWSERSTACK_USERNAME: ${{ secrets.BROWSERSTACK_USERNAME }}
           BROWSERSTACK_ACCESS_KEY: ${{ secrets.BROWSERSTACK_ACCESS_KEY }}
-        run: yarn test:browserstack --log-level ${{ github.run_attempt == 1 && 'info' || 'debug' }}
+        run: yarn test:browserstack --log-level ${{ env.RUNNER_DEBUG == 1 && 'debug' || 'info' }}
```
Finesse, line 44: Did you check it? [It doesn't work](https://github.com/fingerprintjs/fingerprintjs-pro/actions/runs/5343292432/jobs/9686688364#step:4:24). Please make sure it works before requesting a review. 

Please fix in pro too.

Finesse (changes requested):

.github/workflows/test.yml
fpkamp: done, sorry, I got trigger happy thinking the original information was correct

Finesse (approved):

PR #917, opened 2023-06-28
feature: Export more utils

bayotop:


src/index.ts
```diff
@@ -45,3 +45,4 @@ export {
   transformSource, // Not used here but adds only 222 uncompressed (60 compressed) bytes of code
   UnknownSources,
 } from './utils/entropy_source'
+export { withIframe } from './utils/dom'
```
Finesse, line 48: Please add a warning comment used with every other undocumented export

Finesse (changes requested):

src/index.ts
Finesse: It already is, my bad

bayotop: I added it below the warning comment above, which I'd assume includes this export too. Should I add another one regardless, to be sure?

Finesse: I mean this message that you can see in IDE or in `fp.d.ts`:

image

It's added right above the function declaration.

Finesse (approved):

PR #918, opened 2023-06-28
docs(README): remove screen from comparison table

JuroUhlar:
Remove screen size from the comparison table as it is not provided in the Pro response to avoid confusion.
https://fingerprintjs.atlassian.net/browse/INTER-25

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -109,7 +109,7 @@ Full product comparison:
     <tr><td colspan="3"><h4>Core Features</h4></td></tr>
     <tr><td>100% open source</td><td align="center">Yes</td><td align="center">No<sup>1</sup></td></tr>
     <tr><td>Accuracy</td><td align="center">up to 60%</td><td align="center"><b>99.5%</b></td></tr>
-    <tr><td><b>Standard identification signals</b><br/><i>Screen, OS, and device name</i></td><td align="center">✓</td><td align="center">✓</td></tr>
+    <tr><td><b>Standard identification signals</b><br/><i>Browser, operating system</i></td><td align="center">✓</td><td align="center">✓</td></tr>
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -109,7 +109,7 @@ Full product comparison:
     <tr><td colspan="3"><h4>Core Features</h4></td></tr>
     <tr><td>100% open source</td><td align="center">Yes</td><td align="center">No<sup>1</sup></td></tr>
     <tr><td>Accuracy</td><td align="center">up to 60%</td><td align="center"><b>99.5%</b></td></tr>
-    <tr><td><b>Standard identification signals</b><br/><i>Screen, OS, and device name</i></td><td align="center">✓</td><td align="center">✓</td></tr>
+    <tr><td><b>Standard identification signals</b><br/><i>Browser, operating system</i></td><td align="center">✓</td><td align="center">✓</td></tr>
```
Finesse, line 112: Just in case, this is what's used inside, not what's provided in output. The advanced signals below are also not provided in Pro output.

JuroUhlar: Ah, yes, you are right. I am trying to avoid confusion around those things being included in the response, people in Discord are bumping into it. I can see how they can look at this table and assume those are output attributes (because for FPJS, they are). 

I will reverse the proposed change and we can either accept the confusion or tweak the line from `Standard identification signals` to `Standard signals used for identification`, what do you think?

Finesse: Up to you. What about "Input signals" and "Output signals"?

Finesse (approved):

readme.md
Finesse: Looks clearer

PR #921, opened 2023-07-07
feat: Release event loop in font sources

dj-stormtrooper:


src/sources/font_preferences.ts
```diff
@@ -52,13 +52,14 @@ export const presets: Record<string, Preset> = {
  * The "min" and the "mono" (only on Windows) value may change when the page is zoomed in Firefox 87.
  */
 export default function getFontPreferences(): Promise<Record<string, number>> {
-  return withNaturalFonts((document, container) => {
+  return withNaturalFonts(async (document, body) => {
     const elements: Record<string, HTMLElement> = {}
     const sizes: Record<string, number> = {}
+    const container = document.createElement('div')
```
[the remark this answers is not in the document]
Finesse: What is "`body` update"?

Could you please leave this comment in the code.

src/sources/font_preferences.ts
```diff
@@ -52,13 +52,14 @@ export const presets: Record<string, Preset> = {
...
 
     // First create all elements to measure. If the DOM steps below are done in a single cycle,
     // browser will alternate tree modification and layout reading, that is very slow.
-    for (const key of Object.keys(presets)) {
+    await mapWithBreaks(Object.keys(presets), async (key) => {
```
Finesse, line 64: Are you sure `mapWithBreaks` is necessary? There are only 7 presets, and the actions inside are very small.

The callback is not necessary to be `async`, there is no `await` inside, and `mapWithBreaks` is made for synchronous callbacks. The `async` only makes browsers create unnecessary promises at runtime.

src/sources/fonts.ts
```diff
@@ -102,16 +103,16 @@ export default function getFonts(): Promise<string[]> {
...
 
     // creates spans for the fonts to detect and adds them to fontsDiv
-    const initializeFontsSpans = () => {
+    const initializeFontsSpans = async () => {
       // Stores {fontName : [spans for that font]}
       const spans: Record<string, HTMLSpanElement[]> = {}
 
       for (const font of fontList) {
-        spans[font] = baseFonts.map((baseFont) => createSpanWithFonts(font, baseFont))
+        spans[font] = await mapWithBreaks(baseFonts, (baseFont) => createSpanWithFonts(font, baseFont))
```
Finesse, line 115: There are only 3 `baseFonts`, so `mapWithBreaks` should be excessive here. Same on the line 106.

Finesse: It makes sense to use `mapWithBreaks` for the `for (const font of fontList)` loop instead.

Finesse (commented):
Did you measure the performance improvement?

AFAIK browsers don't recalculate the layout until necessary, i.e. `.getBoundingClientRect()` or `.offsetWidth` call in this case. Before that, it shouldn't matter how the elements are created.

Probably, it makes sense to add loop releases before the layout recalculations to let other processes go.

dj-stormtrooper:
> Did you measure the performance improvement?
> 
> AFAIK browsers don't recalculate the layout until necessary, i.e. `.getBoundingClientRect()` or `.offsetWidth` call in this case. Before that, it shouldn't matter how the elements are created.
> 
> Probably, it makes sense to add loop releases before the layout recalculations to let other processes go.

Yes, sure, I noticed strange behaviour after adding `mapWithBreaks` to font preference signal (timings were increased significantly), adding intermediate container improved timings:

```
// sync implementation 
{ median: 62, mean: 63.5, max: 74, min: 59 },
// with async implementation + intermediate container
{ median: 13, mean: 12.7, max: 16, min: 11 }
```

Finesse:
> Yes, sure, I noticed strange behaviour after adding `mapWithBreaks` to font preference signal (timings were increased significantly), adding intermediate container improved timings:
> 
> ```
> // sync implementation 
> { median: 62, mean: 63.5, max: 74, min: 59 },
> // with async implementation + intermediate container
> { median: 13, mean: 12.7, max: 16, min: 11 }
> ```

Looks weird. What about the general performance?

dj-stormtrooper:
> Looks weird

I checked again, look like most of the impact comes from the `fonts` improvements, so we can revert font preferences to keep code simple

> What about the general performance?

It's not super noticeable in case of Fingerprint Pro because we have other bottlenecks, but other async sources that run for ~50ms run faster with released event loop

I attached detailed reports to notion task

src/sources/font_preferences.ts
dj-stormtrooper: I checked both cases, `mapWithBrakes` never reaches `await wait(0)` in tests 

So perhaps durations are only affected by moving sync operations to microtasks

src/sources/fonts.ts
Finesse: I'm concerned about many microtasks because they can harm the performance. What do you think if we just release the loop once before measuring the DOM elements (both in `fonts` and `font_preferences`)?

PR #926, opened 2023-07-20
feat: Release event loop in synchronous signals

dj-stormtrooper:
I measured timings of these sources before the changes were implemented, details are in Notion

### DIFF
--- a/src/sources/canvas.ts
+++ b/src/sources/canvas.ts
@@ -1,3 +1,5 @@
+import { releaseEventLoop } from '../utils/async'
+
 export interface CanvasFingerprint {
   winding: boolean
   geometry: string
@@ -5,7 +7,7 @@ export interface CanvasFingerprint {
 }
 
 // https://www.browserleaks.com/canvas#how-does-it-work
-export default function getCanvasFingerprint(): CanvasFingerprint {
+export default async function getCanvasFingerprint(): Promise<CanvasFingerprint> {
   let winding = false
   let geometry: string
   let text: string
@@ -17,6 +19,7 @@ export default function getCanvasFingerprint(): CanvasFingerprint {
     winding = doesSupportWinding(context)
 
     renderTextImage(canvas, context)
+    await releaseEventLoop()
     const textImage1 = canvasToString(canvas)
     const textImage2 = canvasToString(canvas) // It's slightly faster to double-encode the text image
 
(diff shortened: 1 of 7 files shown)


src/sources/dom_blockers.ts
```diff
@@ -356,6 +356,8 @@ export async function getBlockedSelectors<T extends string>(selectors: readonly
   }
   d.body.appendChild(root)
 
+  await releaseEventLoop()
```
Finesse, line 359: Looks like this may cause the elements to be displayed on the page for a moment. Probably, `root` should be hidden with CSS such way that `offsetParent` keeps working.

dj-stormtrooper: We can skip this source to eliminate the risk, considering low profit of this optimisation

Finesse: Sounds good

Finesse: The other DOM sources may have the same issues. `visibility: hidden` will suffice to solve them.

dj-stormtrooper: Ok, I'll check

Finesse (approved):

dj-stormtrooper:
Please do not merge yet, I want to check stability of these signals, I suspect that it's worse than before

dj-stormtrooper:
Yes, the stability isn't great (30 runs for each test)

image

Let's discard the idea, too many different issues for this task

Finesse:
> Yes, the stability isn't great (30 runs for each test)

This is weird. Was it better before the changes?

dj-stormtrooper:
> > Yes, the stability isn't great (30 runs for each test)
> 
> This is weird. Was it better before the changes?

I think you're right, before the changes the behaviour is similar. It's just harder to notice because we use retries on CI, but locally it's the same

src/sources/canvas.ts
```diff
@@ -1,11 +1,13 @@
...
+
 export interface CanvasFingerprint {
   winding: boolean
   geometry: string
   text: string
 }
 
 // https://www.browserleaks.com/canvas#how-does-it-work
-export default function getCanvasFingerprint(): CanvasFingerprint {
+export default async function getCanvasFingerprint(): Promise<CanvasFingerprint> {
```
Finesse, line 10: Since the source is now asynchronous, please move it to the list of asynchronous sources in `src/sources/index.ts`. Same in Pro agent.

Finesse (changes requested):

src/sources/canvas.ts
dj-stormtrooper: Fixed (I'll update the Pro once we merge this PR)

Finesse (approved):

PR #931, opened 2023-07-30
feat: non-ASCII support and optimization in `murmurhash3`

dj-stormtrooper:
I added proper support for non-ASCII symbols for `murmurhash3_128_x64`. Previous implementation wasn't compatible with implementation in other languages (e.g. Golang, C++) for symbols with charcodes outside of `0 - 0xff` range. Also it was limited in terms of uniqueness, e.g. `hash('Hello, world, hi')` was equal to `hash('ňťŬŬůĬĠŷůŲŬŤĬĠŨũ')`.
Current new implementation are free of these issues

Also I refactored internal functions to achieve minimal minimal memory allocations and reduce GC load, it gained performance boost: 
* Chrome: 27%
* Webkit: 58%
* Firefox: 45%

[Benchmark](https://jsbench.me/p4lkpaoabi/1) (sorry for the minified code in the benchmark, `jsbench` allows to save suite only if fields are below 2048 symbols, so I tried to make it more compact)

<details>
  <summary>JSBench screenshots</summary>

image

Webkit

image

Chrome

image

Firefox
</details>

### DIFF
--- a/src/utils/data.ts
+++ b/src/utils/data.ts
@@ -149,3 +149,42 @@ export function maxInIterator<T>(iterator: Iterator<T>, getItemScore: (item: T) 
 
   return maxItem
 }
+
+function encode(text: string): Uint8Array {
+  // Benchmark: https://jsbench.me/b6klaaxgwq/1
+  // If you want to just count bytes, see solutions at https://jsbench.me/ehklab415e/1
+  if (typeof TextEncoder === 'function') {
+    return new TextEncoder().encode(text) // From https://stackoverflow.com/a/11411402/1118709
+  }
+
+  // From https://stackoverflow.com/a/18722848/1118709
+  const binaryText = unescape(encodeURI(text))
+  const bytes = new Uint8Array(binaryText.length)
+
+  for (let i = 0; i < binaryText.length; ++i) {
+    bytes[i] = binaryText.charCodeAt(i)
+  }
+
+  return bytes
+}
+
+/**
+ * Converts a string to UTF8 bytes
+ *
+ * Warning for package users:
+ * This function is out of Semantic Versioning, i.e. can change unexpectedly. Usage is at your own risk.
+ */
+export function getUTF8Bytes(input: string): Uint8Array {
+  const result = new Uint8Array(input.length)
+  for (let i = 0; i < input.length; i++) {
+    // `charCode` is faster than encoding so we prefer that when it's possible
+    const charCode = input.charCodeAt(i)
+
+    // In case of non-ASCII symbols we use proper encoding
+    if (charCode < 0 || charCode > 127) {
+      return encode(input)
+    }
+    result[i] = charCode
+  }
+  return result
+}
--- a/src/utils/hashing.test.ts
+++ b/src/utils/hashing.test.ts
@@ -26,7 +26,15 @@ const longText =
 
 describe('Murmur3', () => {
   it('makes x64 128 bit hash', () => {
-    expect(x64hash128('Hello, world')).toBe('ebd28b45027ab97477416103e3fff7b8')
+    const input = 'Hello, world, hi'
+    const nonAsciiInput = 'ňťŬŬůĬĠŷůŲŬŤĬĠŨũ' // 'Hello, world, hi' + 256 codepoints for each to make it non-ASCII
+    const shortInput = 'hello'
+
+    expect(x64hash128(input)).not.toBe(x64hash128(nonAsciiInput))
+
+    expect(x64hash128(input)).toBe('9a66b4567d520770dc8eaf9a508ecf1b')
+    expect(x64hash128(nonAsciiInput)).toBe('460892d2cab76edff07f62a97e106f6b')
+    expect(x64hash128(shortInput)).toBe('cbd8a7b341bd9b025b1e906a48ae1d19')
     expect(x64hash128(longText)).toBe('211a6f425b82e115fb52ccdc51edb290')
   })
 })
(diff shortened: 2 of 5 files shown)


src/utils/data.ts
```diff
@@ -149,3 +149,42 @@ export function maxInIterator<T>(iterator: Iterator<T>, getItemScore: (item: T)
...
+
+function encode(text: string): Uint8Array {
+  // Benchmark: https://jsbench.me/b6klaaxgwq/1
+  // If you want to just count bytes, see solutions at https://jsbench.me/ehklab415e/1
+  if (typeof TextEncoder === 'function') {
+    return new TextEncoder().encode(text) // From https://stackoverflow.com/a/11411402/1118709
+  }
+
+  // From https://stackoverflow.com/a/18722848/1118709
+  const binaryText = unescape(encodeURI(text))
```
Finesse, line 161: The fallback is not necessary, all browsers that we support (and some older browsers) do support `TextEncoder`

src/utils/hashing.test.ts
```diff
@@ -26,7 +26,22 @@ const longText =
 
 describe('Murmur3', () => {
   it('makes x64 128 bit hash', () => {
-    expect(x64hash128('Hello, world')).toBe('ebd28b45027ab97477416103e3fff7b8')
+    const input = 'Hello, world, hi'
+    const inputLessThanChunk = 'Hello, world, h'
+    const inputGreaterThanChunk = 'Hello, world, hi!'
+    const inputGreaterThan2Chunks = 'Hello, world, hi, Hello, world, hi'
+
+    const nonAsciiInput = 'ňťŬŬůĬĠŷůŲŬŤĬĠŨũ' // 'Hello, world, hi' + 256 codepoints for each to make it non-ASCII
```
Finesse, line 34: What do you think if this string is computed from `input` instead of hard-coding?

src/utils/hashing.ts
```diff
@@ -2,229 +2,295 @@
  * Taken from https://github.com/karanlyons/murmurHash3.js/blob/a33d0723127e2e5415056c455f8aed2451ace208/murmurHash3.js
```
Finesse, line 2: The new code is heavily modified, so let's change to "Based on"

src/utils/hashing.ts
Finesse, line 1: @Valve Since the code if heavily modified, what do you think if we remove the reference from the code banner?

image

(in Pro agent too)

src/utils/hashing.ts
```diff
@@ -2,229 +2,295 @@
...
+const N2 = [0, 0x38495ab5]
+/**
+ * Given a string and an optional seed as an int, returns a 128 bit
+ * hash using the x64 flavor of MurmurHash3, as an unsigned hex.
+ * All internal functions mutates passed value to achieve minimal memory allocations and GC load
+ *
+ * Benchmark https://jsbench.me/p4lkpaoabi/1
+ */
+export function x64hash128(input: string, seed?: number): string {
+  const key = getUTF8Bytes(input)
```
Finesse, line 162: Just an idea, probably will give a minor or no effect: iterate over the blocks of 16 bytes instead of making a full array of bytes.

Finesse (commented):
Please list shortly the optimizations that you applied in the PR description

src/utils/hashing.ts
Valve: The license is vague on when we should provide the attribution. It says:
"The above copyright notice and this permission notice shall be included in all
copies or substantial portions of the Software".
Merriam Webster says that "substantial" means: "being largely but not wholly that which is specified".
So I would say if we retain less than 50% of the original source verbatim, we can remove the attribution.

Finesse: @dj-stormtrooper Could you please estimate the portion of the original code still present in our code?

src/utils/hashing.ts
dj-stormtrooper: Do you mean to collect bytes while iterating through unicode codepoints until we have 16-bytes block?

Finesse: Yes, because as I can see, the algorithm operates 16-byte blocks

dj-stormtrooper: For history: this was a good exersice but the approach is slower (maybe my custom implementation just can't beat some `TextEncoder` optimizations)

Code points implementation https://github.com/fingerprintjs/fingerprintjs/pull/936

src/utils/hashing.ts
dj-stormtrooper: I checked with this tool https://dolos.ugent.be/

* [Including comments](https://dolos.ugent.be/server/#/share/2728318240659492154)
image

* [Without comments
](https://dolos.ugent.be/server/#/share/4488101566508081510)
image

The comparison also includes variable names (see the diff in the report) but I didn't change them unless it's necessary

dj-stormtrooper:
> Please list shortly the optimizations that you applied in the PR description

Done

src/utils/hashing.ts
Finesse: @dj-stormtrooper Thanks!

@Valve The similarity is small, and [the original repository](https://github.com/karanlyons/murmurHash3.js) seems to be abandoned (the last commit was in 2020). So I suggest removing the attribution from the license banner, and changing the «taken from» references to «based on» or «inspired by». Do you agree?

Finesse: @Valve A kind reminder. We need just "yes" or "no" from you.

Valve: Sure, we can use "inspired by", up to you.

Finesse (approved):

PR #939, opened 2023-08-12
feat: `murmur3hash`  WASM PoC

dj-stormtrooper:
image

---

Bundle size:
```
minified GZIP Current 16K vs 22K Rust

minified JS Current 37K vs 49k Rust

unminified JS Current 135K vs 144K Rust
```

---

P.S.: Unit tests are broken because autogenerated JS-code uses `import.meta` which `karma-typescript` doesn't support, but the algorithm works fine (I checked it manually on same test-cases)

We can fix it by adding additional build step (which we will add if we move it to the separate package) or we can rewrite autogenerated JS-code, produced by `wasm-pack`, and omit unsupported syntax


Finesse:
@dj-stormtrooper What's the status of the PR?

dj-stormtrooper:
I'll measure few additional performance tests (based on size) and we will decide should we move it to the external package and imlpement

wasm-rust/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+description = "A sample project with wasm-pack"
+license = "MIT/Apache-2.0"
+edition = "2018"
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+mur3 = "0.1.0"
+wasm-bindgen = "0.2"
```
GoldStrikeArch, line 14: Comment to wasm-bindgen for future reference

PR #942, opened 2023-08-29
Add a function to detect Safari 17

Finesse:
Unfortunately, BrowserStack doesn't support Safari 17 in Automation yet (both macOS and iOS), so the test doesn't run in Karma. I've checked it manually.

### DIFF
--- a/src/utils/browser.ts
+++ b/src/utils/browser.ts
@@ -99,6 +99,7 @@ export function isWebKit(): boolean {
  */
 export function isDesktopSafari(): boolean {
   const w = window
+  const { HTMLElement, Document } = w
 
   return (
     countTruthy([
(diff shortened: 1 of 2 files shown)


src/utils/browser.ts
```diff
@@ -99,15 +99,16 @@ export function isWebKit(): boolean {
...
   return (
     countTruthy([
       'safari' in w, // Always false in Karma and BrowserStack Automate
       !('ongestureend' in w),
       !('TouchEvent' in w),
       !('orientation' in w),
-      'HTMLElement' in w && !('autocapitalize' in HTMLElement.prototype),
-      'Document' in w && 'pointerLockElement' in Document.prototype,
+      HTMLElement && !('autocapitalize' in HTMLElement.prototype),
+      Document && 'pointerLockElement' in Document.prototype,
```
Finesse, line 111: This change is just to make the minified code a bit smaller

bayotop (approved):

PR #949, opened 2023-09-15
feature: Add browser-based port scanning

basvdlouw:
Browser based port scanning can be used as a fingerprinting technique by detecting open HTTP ports, leveraging the no-cors mode in the Fetch API. Obviously, the more ports are scanned, the more time it will take to complete the scan, I have currently implemented 10000 ports which takes roughly 10 seconds.

### DIFF
--- a/src/sources/index.ts
+++ b/src/sources/index.ts
@@ -1,4 +1,5 @@
 import { loadSources, SourcesToComponents } from '../utils/entropy_source'
+import getOpenPorts from './open_ports'
 import getAudioFingerprint from './audio'
 import getFonts from './fonts'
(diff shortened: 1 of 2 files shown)


Finesse:
Hey @basvdlouw. I'm sorry, but we won't add port scanning to FingerprintJS for these reasons:

1. This is an intrusive technique leaving many traces in browser dev tools and looking too impudent. This is beyond the concept of fingerprinting.
2. It takes too much time. Fingerprinting is expected to take not longer than a second.

If you can't solve these, I'll have to close the PR.

basvdlouw:
Hi, makes sense. I can indeed not solve these

PR #953, opened 2023-09-20
Simplify append elements in withNaturalFonts

Connormiha:
Supported browsers has `append` method, which is more convenient.


Finesse:
@Connormiha Looks good to me. Could you please use `append` in the other places where `appendChild` is called multiple times a row on the same element?

Connormiha:
> @Connormiha Looks good to me. Could you please use `append` in the other places where `appendChild` is called multiple times a row on the same element?

Done

Finesse (approved):

Finesse:
Thank you

PR #954, opened 2023-09-26
Simplified checking for equal sets

Connormiha:
I've simplified the hard-to-read loop a bit.

UPD: removed support of old browsers.

### DIFF
--- a/src/utils/data.ts
+++ b/src/utils/data.ts
@@ -115,17 +115,12 @@ export function areSetsEqual(set1: Set<unknown>, set2: Set<unknown>): boolean {
     return false
   }
 
-  if (set1.values) {
-    for (let iter = set1.values(), step = iter.next(); !step.done; step = iter.next()) {
-      if (!set2.has(step.value)) {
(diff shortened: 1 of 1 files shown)


src/utils/data.ts
```diff
@@ -124,13 +124,7 @@ export function areSetsEqual(set1: Set<unknown>, set2: Set<unknown>): boolean {
...
     // An implementation for browsers that don't support Set iterators
-    let areEqual = true
-    set1.forEach((value) => {
-      if (areEqual && !set2.has(value)) {
-        areEqual = false
-      }
-    })
-    return areEqual
+    return set1.every((value) => set2.has(value));
   }
```
Finesse, line 128: This is not necessary anymore, because all supported browsers support `values`. So let's remove the condition, leaving only the code in the positive branch.

Finesse: BTW `Set` doesn't have `every` method

Finesse (approved):

PR #983, opened 2024-01-26
fix: WebGL console warning in Chrome 121 and Safari 17.4

Finesse:
This remove an unwanted console message arising in:

- Blink (Chrome) 121: `webgl.ts:139 WebGL: this extension has very low support on mobile devices; do not rely on it for rendering effects: WEBGL_polygon_mode`
- WebKit (Safari) 17.4: `WebGL: non-portable extension enabled: WEBGL_polygon_mode`

The PR disables the `WEBGL_polygon_mode` extension, but this is not a fingerprint-breaking change, because the extension didn't exist in the previous versions. Though the browser updates change the fingerprint because of other changes to WebGL.

### DIFF
--- a/src/sources/webgl.ts
+++ b/src/sources/webgl.ts
@@ -1,4 +1,4 @@
-import { isGecko } from '../utils/browser'
+import { isChromium, isGecko, isWebKit } from '../utils/browser'
 
 // Types and constants are used instead of interfaces and enums to avoid this error in projects which use this library:
 // Exported variable '...' has or is using name '...' from external module "..." but cannot be named.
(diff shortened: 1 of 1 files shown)


src/sources/webgl.ts
```diff
@@ -132,7 +133,8 @@ export function getWebGlExtensions({ cache }: Options): WebGlExtensionsPayload |
   // Extension parameters
   if (extensions) {
     for (const name of extensions) {
-      if (name === rendererInfoExtensionName && shouldAvoidDebugRendererInfo()) {
+      // The "polygon mode" extension causes a console warning in Chromium and WebKit
```
Finesse, line 136: Firefox emits no warning because it doesn't support `WEBGL_polygon_mode` yet (even the nightly 124 version). When it does support, it's very likely to have a similar warning too.

dj-stormtrooper (approved):

PR #989, opened 2024-02-23
Allowing playground to be run on a nominated interface and port

david-dick:


Finesse:
This is already possible:

```bash
yarn playground:start --port 7777 --host 192.168.0.12
```

Considering this, are your changes necessary?

PR #990, opened 2024-02-26
Bugfix: values of canvas geometry and text returned in reverse order.

shbobur:
Values of geometry and text in canvas are reversed. 

The demo site confirms it: https://fingerprintjs.github.io/fingerprintjs/
I copy/paste geometry and text values to https://codebeautify.org/base64-to-image-converter, and can see the images are swapped.


Finesse:
Nice catch, thank you

Finesse (approved):

PR #1000, opened 2024-04-15
update debug expression in test workflow

fpkamp:


Finesse:
Is it necessary? As I can see, [it works](https://github.com/fingerprintjs/fingerprintjs/actions/runs/8450699083/job/23814973939#step:9:57) before the PR.

PR #1004, opened 2024-05-23
feat: AudioContext baseLatency entropy source

dj-stormtrooper:


src/sources/index.ts
```diff
@@ -92,7 +93,7 @@ export const sources = {
   architecture: getArchitecture,
   applePay: getApplePayState,
   privateClickMeasurement: getPrivateClickMeasurement,
-
+  audioBaseLatency: getAudioContextBaseLatency,
```
Finesse, line 96: Let's keep the blank line before the comment

src/sources/audio_base_latency.test.ts
```diff
@@ -0,0 +1,38 @@
...
+})
+
+function hasBaseLatencySupport() {
+  if (isWebKit()) {
+    const { major, minor } = getBrowserVersion() || { major: 0, minor: 0 }
+    if (major < 14) {
+      return false
+    } else {
+      return minor >= 1
+    }
```
Finesse, line 35: Looks like the condition is incorrect. Versions 15.0, 16.0 and so on will give `false`.

src/sources/audio_base_latency.test.ts
```diff
@@ -0,0 +1,38 @@
+import { getBrowserVersion, isGecko, isWebKit } from '../../tests/utils'
+import getAudioBaseLatency from './audio_base_latency'
+
+describe('Sources', () => {
+  describe('audio base latency', () => {
```
Finesse, line 5: We use camelCase source names in these tests

src/sources/audio_base_latency.ts
```diff
@@ -0,0 +1,7 @@
+export default function getAudioContextBaseLatency(): number | undefined {
+  if (!window.AudioContext) {
+    return undefined
+  }
+
+  return new AudioContext().baseLatency
```
bayotop, line 6: nit: Could we reuse the AudioContext() from the existing audio signal?

dj-stormtrooper: It looks that we can't, In existing signal we use `OfflineAudioContext` which doesn't have this property

Let me know if I'm missing something

Finesse (approved):

PR #1005, opened 2024-05-28
Update ApplePay exports

bayotop:


src/sources/apple_pay.ts
Finesse, line 1: Let's also try to replace
```js
export function getStateFromError(error: unknown): -2 | -3 {
```
with
```js
function getStateFromError(error: unknown): ApplePayState {
```

And we don't need the "Warning for package users" paragraph in the comment.

PR #1024, opened 2024-08-27
Fix the "detects Safari" test failing in Safari 15

Finesse:


src/utils/browser.ts
```diff
@@ -121,27 +121,21 @@ export function isDesktopWebKit(): boolean {
...
-  if (!isFunctionNative(w.print)) {
-    return false // Chrome, Firefox, Yandex, DuckDuckGo macOS, Edge
-  }
-
   return (
-    countTruthy([
-      // Incorrect in Safari <= 14 (iOS and macOS)
-      String((w as unknown as Record<string, unknown>).browser) === '[object WebPageNamespace]',
-      // Incorrect in desktop Safari and iOS Safari <= 15
-      'MicrodataExtractor' in w,
```
Finesse, line 143: I've removed this condition because it works only in Safari 17

bayotop (approved):

PR #1027, opened 2024-08-30
feat: Add `unsupported` field to the `webGlExtensions` entropy source

xnerhu:


src/sources/webgl.ts
```diff
@@ -24,6 +24,7 @@ type WebGlExtensionsPayload = {
   shaderPrecisions: string[] // ['FRAGMENT_SHADER.LOW_FLOAT=127,127,23...
   extensions: string[] | null // ['ANGLE_instanced_arrays', 'EXT_blend_minmax', 'EXT_color...
   extensionParameters: string[] // ['COMPRESSED_RGB_S3TC_DXT1_EXT(33776)', 'COMPR...
+  supported: boolean[]
```
Finesse, line 27: - The name should have more context. For example `supportedExtensions`.
- The fact that this field references to the `extensions` field is not obvious, so this should be mentioned in a comment
- Because the array is all-true in almost all cases, it makes sense to keep only unsupported extensions. It will reduce the payload size. This can be either extension names, or the `extensions`'s indices (with a comment telling about the reference)

src/sources/webgl.ts
```diff
@@ -133,15 +135,19 @@ export function getWebGlExtensions({ cache }: Options): WebGlExtensionsPayload |
...
+      supported.push(extensionSupported)
+
       if (
         (name === rendererInfoExtensionName && shouldAvoidDebugRendererInfo()) ||
         (name === polygonModeExtensionName && shouldAvoidPolygonModeExtensions())
       ) {
         continue
       }
 
-      const extension = gl.getExtension(name)
```
Finesse, line 143: This must be called after the above checks. Please read the comments for `shouldAvoidDebugRendererInfo` and `shouldAvoidPolygonModeExtensions` to know why. If you want to check `WEBGL_debug_renderer_info` and `WEBGL_polygon_mode` specifically, it's not doable.

Finesse (changes requested):

src/sources/webgl.ts
```diff
@@ -24,6 +24,7 @@ type WebGlExtensionsPayload = {
   shaderPrecisions: string[] // ['FRAGMENT_SHADER.LOW_FLOAT=127,127,23...
   extensions: string[] | null // ['ANGLE_instanced_arrays', 'EXT_blend_minmax', 'EXT_color...
   extensionParameters: string[] // ['COMPRESSED_RGB_S3TC_DXT1_EXT(33776)', 'COMPR...
+  unsupportedExtensions: string[]
```
Finesse, line 27: Please add a note like for the other fields.

In general, try to keep the code you add at least as good as the code around.

Finesse (commented):
Looks good. Let's keep unmerged until we agree on the Fingerprint Pro part of the changes.

Finesse (approved):

PR #1054, opened 2024-10-28
NPM publishing workflow

Finesse:
The PR adds a GitHub Actions workflow that starts manually, publishes the current version to NPM and creates a corresponding Git tag. The workflow can run only on the `master` branch (until somebody changes the environment settings in the repository settings).

The goal of the workflow is allowing the repository maintainers to publish the loader package to NPM with having no NPM access token. Limiting the workflow to `master` ensures that at least 2 people are needed to publish a new version to NPM.

### DIFF
--- a/.github/workflows/npm_publish.yml
+++ b/.github/workflows/npm_publish.yml
@@ -4,14 +4,14 @@
 # The environment should be protected to allow running only on the master branch.
 
 name: Publish to NPM
-on: [workflow_dispatch, push] # todo: Remove the temporary "push" trigger
+on: workflow_dispatch
 
 jobs:
   publish:
     name: Build and publish
     runs-on: ubuntu-latest
     timeout-minutes: 5
-    # environment: npm # todo: Uncomment
+    environment: npm
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


.github/workflows/npm_publish.yml
```diff
@@ -0,0 +1,49 @@
...
+          tag=$(echo "$version" | grep -oP '(?<=-)[^.]+' || echo "latest")
+          echo -e "Version: $version\nVersion tag: $tag"
+          echo -e "version=$version\ntag=$tag" >> $GITHUB_OUTPUT
+      - name: Publish to NPM
+        # This command will automatically fail because of some common mistakes (this is a wanted behavior):
+        # - The current version is already published
+        # - The package uses a local directory as a dependency
+        # - The NPM access token is invalid or expired
+        run: |
+          echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_ACCESS_TOKEN }}" > ~/.npmrc
```
Valve, line 43: Is it possible to read it later to display the access token through the GH action?

Finesse: No, GitHub hides the secrets in the job output automatically. Moreover, this secret is available only on the `master` branch, so one needs to make a PR and get an approval before they can use the secret differently. This is an example of what the job output will look like:

image

Valve: this is great, thanks!

spalt08 (approved):

PR #1059, opened 2024-12-09
Add a `withIframe` test for handling delayed document.body availability

primarch42:
This test ensures that the withIframe function properly handles scenarios where document.body is initially unavailable, simulating delayed document loading. It verifies that the promise remains pending until document.body becomes available and resolves correctly once it is accessible.

### DIFF
--- a/src/utils/dom.test.ts
+++ b/src/utils/dom.test.ts
@@ -18,23 +18,25 @@ describe('DOM utilities', () => {
     expect(isAnyParentCrossOrigin()).toBeFalse()
   })
 
-  describe('Document loading - consistency test', () => {
-    let result: Promise<unknown> | null = null
-    beforeEach(() => jasmine.clock().install())
-    afterEach(() => jasmine.clock().uninstall())
+  describe('withIframe', () => {
+    describe('with initial null document.body', () => {
+      let result: Promise<unknown> | null = null
+      beforeEach(() => jasmine.clock().install())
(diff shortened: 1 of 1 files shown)


src/utils/dom.test.ts
```diff
@@ -16,4 +17,24 @@ describe('DOM utilities', () => {
     // Assuming the tests run only with Karma, which doesn't run tests inside an iframe
     expect(isAnyParentCrossOrigin()).toBeFalse()
   })
+
+  describe('Document loading - consistency test', () => {
```
Finesse, line 21: Jasmine concatenates all the ancestor description into 1 sentence.

Please also add a describe level telling that it tests withIframe.

For example:

```js
describe('DOM utilities', () => {
  describe('withIframe', () => {
    describe('with null document.body', () => {
      it("doesn't fail", () => {
```
The resulting test sentence will be "DOM utilities withIframe with null document.body doesn't fail".

Finesse (changes requested):

src/utils/dom.test.ts
Finesse: Sorry, the current nesting is excessive. Could you please do like this:

```js
  describe('withIframe', () => {
    it("doesn't fail when document loading is delayed", async () => {
      try {
        jasmine.clock().install()

        // The actual test

      } finally {
        jasmine.clock().uninstall()
      }
    })
  })
```

GoldStrikeArch (approved):

PR #1070, opened 2025-01-22
Add window.Intl.DateTimeFormat().resolvedOptions().locale as a new signal

primarch42:
**Description:**

This PR introduces support for collecting the `locale` value from `window.Intl.DateTimeFormat().resolvedOptions()` as a new signal. Currently, we only collect the `timeZone` signal from this API, but the `locale` value can provide additional valuable insights into user preferences and browser configurations.

### Key Changes:
1. **New Signal Added:**
   - Added logic to capture `window.Intl.DateTimeFormat().resolvedOptions().locale` as a new signal.
   - This signal complements the existing `timeZone` signal and provides more context about the user's environment.

2. **Browser Compatibility:**
   - The `Intl.DateTimeFormat` API is widely supported across modern browsers, making this a low-risk addition.
   - Fallback mechanisms are in place to handle cases where the locale value is not available or `window.Intl` is not supported

### Testing:
- The changes have been tested across multiple browsers and environments to ensure consistent behavior.
- Integration tests have been added to verify that the `locale` signal is correctly captured and handled.

### DIFF
--- a/src/sources/date_time_locale.test.ts
+++ b/src/sources/date_time_locale.test.ts
@@ -1,14 +1,5 @@
 import getDateTimeLocale from './date_time_locale'
 
-// This interface is not exported from lib.es5.d.ts, so it has been copy/pasted here to allow referencing it during
-// the initialization of the spy and to satisfy the TypeScript definition.
-interface ResolvedDateTimeFormatOptions {
-  locale: string
-  calendar: string
-  numberingSystem: string
-  timeZone: string
-}
-
 describe('Sources', () => {
   describe('dateTimeLocale', () => {
     describe('cases for browsers that have support for DateTimeFormat locale', () => {
@@ -30,6 +21,7 @@ describe('Sources', () => {
       afterEach(() => {
         window.Intl = originalIntl
       })
+
       it('should return an empty string if Intl.DateTimeFormat is not supported', () => {
         window.Intl = undefined as unknown as typeof Intl
 
@@ -39,7 +31,7 @@ describe('Sources', () => {
 
       it('should return an empty string if resolvedOptions().locale is undefined', () => {
         spyOn(window.Intl, 'DateTimeFormat').and.returnValue({
-          resolvedOptions: () => ({ locale: undefined } as unknown as ResolvedDateTimeFormatOptions),
+          resolvedOptions: () => ({ locale: undefined } as unknown as Intl.ResolvedDateTimeFormatOptions),
           format: () => '',
           formatToParts: () => [],
         })
--- a/src/sources/date_time_locale.test.ts
+++ b/src/sources/date_time_locale.test.ts
@@ -11,7 +11,7 @@ interface ResolvedDateTimeFormatOptions {
 
 describe('Sources', () => {
   describe('dateTimeLocale', () => {
-    describe('cases for browsers that support Intl or locale', () => {
+    describe('cases for browsers that have support for DateTimeFormat locale', () => {
       it('returns string representing dateTime locale', () => {
         const result = getDateTimeLocale()
         expect(typeof result).toBe('string')
--- a/src/sources/date_time_locale.test.ts
+++ b/src/sources/date_time_locale.test.ts
@@ -1,0 +1,12 @@
+import getDateTimeLocale from './date_time_locale'
+
+describe('Sources', () => {
+  describe('dateTimeLocale', () => {
+    it('returns string representing dateTime locale', () => {
+      const result = getDateTimeLocale()
+      expect(typeof result).toBe('string')
+      // The combinations returned as results for the current test suite specified browsers are: "en", "en-US", "en-GB".
+      expect(result).toMatch(/^en(-[A-Z]{2})?$/)
+    })
+  })
+})
--- a/src/sources/date_time_locale.test.ts
+++ b/src/sources/date_time_locale.test.ts
@@ -6,9 +6,17 @@ describe('Sources', () => {
       it('returns string representing dateTime locale', () => {
         const result = getDateTimeLocale()
         expect(typeof result).toBe('string')
-        // The combinations returned as results for the current test suite specified browsers are:
-        // "en", "en-US", "en-GB".
-        expect(result).toMatch(/^en(-[A-Z]{2})?$/)
+        // The combinations returned as results for the current test suite specified browsers on the BrowserStack are:
+        // "en", "en-US", "en-GB",
+        // However, we need to keep in mind that this is an open-source codebase, and developers who contribute might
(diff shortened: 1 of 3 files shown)


src/sources/date_time_locale.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/resolvedOptions
+ */
+export default function getDateTimeLocale(): string {
+  const DateTimeFormat = window.Intl?.DateTimeFormat
+  if (DateTimeFormat) {
+    const locale = new DateTimeFormat().resolvedOptions().locale
```
GoldStrikeArch, line 7: maybe this should be simplified like `return new DateTimeFormat().resolvedOptions.locale || ''`

primarch42: it's neat 1-liner, but drastically reduces readability.

I will reorganize code a little bit.

src/sources/date_time_locale.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/resolvedOptions
+ */
+export default function getDateTimeLocale(): string {
+  return getLocaleFromResolvedOption()
+}
+
+function getLocaleFromResolvedOption(): string {
+  return window.Intl?.DateTimeFormat().resolvedOptions().locale || ''
```
GoldStrikeArch, line 9: I am thinking if whether should add more optional chaining just to be safe (maybe there are browsers which do not have `resolvedOptions` function on `DateTimeFormat`. I am leaving it up to you but I would do it just to be sure.

primarch42: I am relying on the state of current implementation of the getTimezone function @ `src/sources/timezone.ts`, please check it... 

It does not necessarily mean that such an implementation ensures reliability, but I believe that within about 3 years, someone would have certainly reported an issue if they had detected it in some browser/environment.

GoldStrikeArch: Got it, then everything is fine.

GoldStrikeArch (dismissed):

src/sources/date_time_locale.test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+describe('Sources', () => {
+  describe('dateTimeLocale', () => {
+    describe('cases for browsers that have support for DateTimeFormat locale', () => {
+      it('returns string representing dateTime locale', () => {
+        const result = getDateTimeLocale()
+        expect(typeof result).toBe('string')
+        // The combinations returned as results for the current test suite specified browsers are:
+        // "en", "en-US", "en-GB".
+        expect(result).toMatch(/^en(-[A-Z]{2})?$/)
```
GoldStrikeArch: Because this is OSS (anyone can contribute) this line might fail if someone with a different default locale will run this test locally (like someone from china, for example). So, I suggest to either wrap this like this

```js
if (process.env.CI) { // or whetever we can do to check if we are running this in browserstack
        expect(result).toMatch(/^en(-[A-Z]{2})?$/)
} 
```
Or modify it so it would check for all possible locales and see if our locale is within that list. Something like this

```js
const validLocales = ["en", "en-US", "en-GB", "ru-RU", "ko-KR", ...] // all possible values
...
expect(validLocales.includes(result))
```

GoldStrikeArch (dismissed):

src/sources/date_time_locale.test.ts
Finesse: There is a similar output in the `languages` entropy source, and the test is very simple there. I think a simple regex will suffice.

src/sources/date_time_locale.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/resolvedOptions
+ */
+export default function getDateTimeLocale(): string {
+  return getLocaleFromResolvedOptions()
+}
```
Finesse, line 6: Is there a reason for this function nesting?

src/sources/date_time_locale.test.ts
```diff
@@ -0,0 +1,52 @@
...
+        // that our logic for detecting local test runs is flexible and inclusive, ensuring it works for all
+        // contributors regardless of their setup.
+        // Examples of locales: "en-US", "zh-Hans-CN", "fr-CA", "es-419", regex formed by following->
+        // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/resolvedOptions#:~:text=locale-,The%20BCP%2047,-language%20tag%20for
+        // https://datatracker.ietf.org/doc/html/rfc4647#section-3.4
+        const localeRegex = /^[a-z]{2,3}(-[A-Za-z]{4})?(-([A-Z]{2}|[0-9]{3}))?(-[A-Za-z0-9]+)*$/
+        expect(result).toMatch(localeRegex)
+      })
+    })
+    describe("edge cases for browsers that don't support Intl or locale", () => {
```
Finesse, line 22: Nit: please leave blank lines between `it`s and `describe`s

src/sources/date_time_locale.test.ts
GoldStrikeArch: Yeah, the current regex `/^[a-z]{2,3}(-[A-Za-z]{4})?(-([A-Z]{2}|[0-9]{3}))?(-[A-Za-z0-9]+)*$/` which follows https://datatracker.ietf.org/doc/html/rfc4647#section-3.4 works well.

src/sources/date_time_locale.ts
primarch42: not at all... I will one-line it

src/sources/date_time_locale.test.ts
primarch42: @Finesse Are you sure about this?

other *.test.ts files have different format
Screenshot 2025-01-26 at 19 53 29
Screenshot 2025-01-26 at 19 53 40

I see blank line only between `it`s;  Additionally linter is forcing it to be just in new line after describe, without new line:
Screenshot 2025-01-26 at 19 56 58

primarch42: done

Finesse (approved):
Please check the integration with Fingerprint Pro before merging

bayotop (approved):

GoldStrikeArch (approved):

PR #1072, opened 2025-01-29
Adjust audio signal tests for Samsung Browser >= 26

GoldStrikeArch:
**Description:**

This PR fixes the `Check upcoming browser versions` workflow. You can see the failed run [here](https://github.com/fingerprintjs/fingerprintjs/actions/runs/13015627749).

### Key Changes:
   - Added a check for Samsung Browser 26 or newer in `doesBrowserPerformAntifingerprinting` function.

PR #1073, opened 2025-01-29
Enforce padding between function definitions with ESLint rule

primarch42:
There is an ESLint rule that defines the desired linting outcome for "Enforce padding between function definitions." 
I believe it's desirable to have this added to avoid similar [issues](https://github.com/fingerprintjs/fingerprintjs/pull/1072/files#r1934181801) with "human-reviewer (not) linting the code" in the future.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -32,11 +32,7 @@
       ],
       "rules": {
         "max-len": ["error", { "code": 120, "ignoreUrls": true }],
-        "@typescript-eslint/no-unused-vars": ["error", { "ignoreRestSiblings": true }],
-        "padding-line-between-statements": [
-          "error",
-          { "blankLine": "always", "prev": "function", "next": "function" }
-        ]
+        "@typescript-eslint/no-unused-vars": ["error", { "ignoreRestSiblings": true }]
       }
     }
   ]
--- a/.eslintrc
+++ b/.eslintrc
@@ -12,7 +12,11 @@
   "parser": "@typescript-eslint/parser",
   "rules": {
     "no-console": "warn",
-    "max-len": ["error", { "code": 120, "ignoreUrls": true }]
+    "max-len": ["error", { "code": 120, "ignoreUrls": true }],
+    "padding-line-between-statements": [
+      "error",
+      { "blankLine": "always", "prev": "function", "next": "function" }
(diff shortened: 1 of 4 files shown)


.eslintrc
```diff
@@ -12,7 +12,11 @@
   "parser": "@typescript-eslint/parser",
   "rules": {
     "no-console": "warn",
-    "max-len": ["error", { "code": 120, "ignoreUrls": true }]
+    "max-len": ["error", { "code": 120, "ignoreUrls": true }],
+    "padding-line-between-statements": [
```
GoldStrikeArch, line 16: I don't think we need to put it here if we are putting it in `overrides` section.

.eslintrc
```diff
@@ -28,7 +32,11 @@
       ],
       "rules": {
         "max-len": ["error", { "code": 120, "ignoreUrls": true }],
-        "@typescript-eslint/no-unused-vars": ["error", { "ignoreRestSiblings": true }]
+        "@typescript-eslint/no-unused-vars": ["error", { "ignoreRestSiblings": true }],
+        "padding-line-between-statements": [
+          "error",
+          { "blankLine": "always", "prev": "function", "next": "function" }
```
GoldStrikeArch, line 38: I am thinking that maybe we should also add this rule for things like `switch` and `while`? Because right now we can write a code like this:
```ts
function foo() {
let run = true;
while(run) { // << no blank line here and it is fine by eslint.
   ...
 }
}
```

.eslintrc
primarch42: removed

.eslintrc
primarch42: Intention was to add rule that targets exclusively blank line between functions.
We can extend rules further outside this PR.

GoldStrikeArch: Yeah, let's extend it in further PRs

GoldStrikeArch (approved):

Finesse:
@GoldStrikeArch Please implement the same change in other agent-relater repositories that use the same ESLint configuration

GoldStrikeArch:
@Finesse 

> Please implement the same change in other agent-relater repositories that use the same ESLint configuration
Sure, no problem.

PR #1074, opened 2025-02-05
DateTimeFormat locale source enhancement

primarch42:
### **Description:**

This PR enhances the existing logic for collecting the locale signal from `window.Intl.DateTimeFormat().resolvedOptions()`. It introduces a mechanism to handle various edge cases where the `Intl` API or the `DateTimeFormat` constructor is not available, or when the `locale` value is undefined, null, or an empty string. These changes provide better compatibility and reliability in environments with incomplete or missing internationalization support.

---

### **Key Changes:**

1. **Enhanced Locale Signal Logic:**
   - The function `getDateTimeLocale` now returns either a valid locale string or a special error indicator from the `SpecialFingerprint` enum.
   - It performs multiple checks to determine whether `Intl` and `DateTimeFormat` are supported and if the locale value is valid.
     - If the `Intl` API is not supported, the function returns `SpecialFingerprint.IntlApiNotSupported`.
     - If the `DateTimeFormat` constructor is missing, it returns `SpecialFingerprint.DateTimeFormatNotSupported`.
     - If the locale is undefined or null, the function returns `SpecialFingerprint.LocaleNotAvailable`.
     - If the locale is an empty string, it's returned as empty string

---

2. **New `SpecialFingerprint` Enum:**

   - Introduced the `SpecialFingerprint` enum to handle special cases related to the locale signal:
```typescript
     export const enum Status {
       /** The browser doesn't support Intl API */
       IntlAPINotSupported = -1,
       /** The browser doesn't support DateTimeFormat constructor */
       DateTimeFormatNotSupported = -2,
       /** DateTimeFormat locale is undefined or null */
       LocaleNotAvailable = -3,
     }
```
---

3. **Cross-Browser Testing:**
   - The changes were validated on multiple browsers and environments to ensure consistent and reliable behavior.

### DIFF
--- a/src/sources/date_time_locale.test.ts
+++ b/src/sources/date_time_locale.test.ts
@@ -1,9 +1,9 @@
-import getDateTimeLocale from './date_time_locale'
+import getDateTimeLocale, { Status } from './date_time_locale'
 
 describe('Sources', () => {
   describe('dateTimeLocale', () => {
     describe('cases for browsers that have support for DateTimeFormat locale', () => {
-      it('returns string representing dateTime locale', () => {
+      it('should return string representing DateTimeFormat locale', () => {
         const result = getDateTimeLocale()
         expect(typeof result).toBe('string')
         // The combinations returned as results for the current test suite specified browsers on the BrowserStack are:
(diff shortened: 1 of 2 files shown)


src/sources/date_time_locale.ts
```diff
@@ -1,6 +1,31 @@
...
+  /** DateTimeFormat locale is undefined or null */
+  LocaleNotAvailable = -3,
+}
+
 /**
  * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/resolvedOptions
  */
-export default function getDateTimeLocale(): string {
-  return window.Intl?.DateTimeFormat().resolvedOptions().locale || ''
+export default function getDateTimeLocale(): string | SpecialFingerprint {
```
Finesse, line 13: Please use number literals instead of const enum in the return type. See [this comment](https://github.com/fingerprintjs/fingerprintjs/blob/374aa1ceb998a3a4a379cb06225c2a40b7f8eebb/src/sources/apple_pay.ts#L20) why.

src/sources/date_time_locale.test.ts
```diff
@@ -31,20 +31,37 @@ describe('Sources', () => {
         window.Intl = originalIntl
       })
 
-      it('should return an empty string if Intl.DateTimeFormat is not supported', () => {
+      it('should return IntlApiNotSupported signal when window.Intl is not available', () => {
```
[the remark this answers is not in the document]
Finesse: It's ok to use "signal" in this repository, but it should be switched everywhere to keep the consistency

GoldStrikeArch (dismissed):

src/sources/date_time_locale.ts
Finesse: Please leave a similar comment here

primarch42: done

Finesse (approved):

GoldStrikeArch (approved):

PR #1093, opened 2025-04-08
feat: handle special 'Inifinty' case in audioBaseLatency signal

GoldStrikeArch:


bayotop (approved):

src/sources/audio_base_latency.test.ts
```diff
@@ -28,6 +28,30 @@ describe('Sources', () => {
...
+
+      const OriginalAudioContext = window.AudioContext
+
+      class MockAudioContext {
+        get baseLatency() {
+          return Infinity
+        }
+      }
+
+      window.AudioContext = MockAudioContext as typeof window.AudioContext
```
primarch42, line 46: Just inline directly:
```ts
...
const OriginalAudioContext = window.AudioContext

window.AudioContext = {
   baseLatency: () => Infinity,
} as unknown as typeof window.AudioContext
...
```

no need for class definition at all

src/sources/audio_base_latency.ts
```diff
@@ -18,5 +20,16 @@ export default function getAudioContextBaseLatency(): number {
   if (!window.AudioContext) {
     return SpecialFingerprint.NotSupported
   }
-  return new AudioContext().baseLatency ?? SpecialFingerprint.NotSupported
+
+  const latency = new AudioContext().baseLatency
+
+  if (latency === null || latency === undefined) {
```
primarch42, line 26: NIT:
"idiom" `==`
```ts
    if (latency == null) { // covers both null & undefined
    ...
    }
```

No other value (like 0, '', false, NaN, etc.) is equal to `null` or `undefined` with `==`

PR #1102, opened 2025-07-11
Chore(build): Set compilation target to ES2018 & remove tslib dependency

primarch42:
This pull request introduces two build-related changes aimed at simplifying the codebase and reducing unnecessary dependencies and output size of produced bundles.

### Changes:

**Updated TypeScript compilation target to ES2018**
The project now compiles directly to ES2018, which is a well-supported baseline for modern browsers. This target provides native support for:
- async/await
- Object.entries, Object.values
- for...of on built-ins
...

This eliminates the need to "downlevel" modern JavaScript features, reducing output size (and complexity).

**Removed tslib dependency**
- Even with `tslib` (still) installed and `importHelpers: true`, when targeting ES2018, the final bundles no longer include references to `tslib`, as the emitted code relies solely on native JavaScript features.
- Since tslib is removed and `importHelpers` option is now disabled and our compilation target is ES2018, the TypeScript compiler no longer injects helper functions that require tslib.

final

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,9 +44,6 @@
     "check:dts": "tsc --isolatedModules --noEmit dist/fp.d.ts",
     "check:ssr": "node --require './dist/fp.cjs.js' --eval '' || (echo \"The distributive files can't be used with server side rendering. Make sure the code doesn't use browser API until an exported function is called.\" && exit 1)"
   },
-  "dependencies": {
-    "tslib": "^2.4.1"
-  },
   "devDependencies": {
     "@fpjs-incubator/broyster": "^0.2.3",
     "@rollup/plugin-json": "^5.0.1",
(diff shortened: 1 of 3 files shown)


GoldStrikeArch (approved):

Finesse:
> * Even with `tslib` (still) installed and `importHelpers: true`, when targeting ES2018, the final bundles no longer include references to `tslib`, as the emitted code relies solely on native JavaScript features.
> * Since tslib is removed and `importHelpers` option is now disabled and our compilation target is ES2018, the TypeScript compiler no longer injects helper functions that require tslib.

Even though no helpers are required now, they can become required in future. That will cause the TS compiler to inject the helpers into every file using it, i.e. create a code duplication. To fix that, you can keep `importHelpers` and make `tslib` a dev-dependency instead, i.e. cause the helpers to be injected into the compiled code once.

This is low-chance and low-impact outcome, so it's ok to leave the changes as is.

PR #1155, opened 2026-04-12
fix: eliminate XSS vector by replacing innerHTML with programmatic DOM construction

### LINKED ISSUE #1154: Security: XSS vector via unsanitized innerHTML in playground
## Description

The `textToDOM` function in `playground/index.ts` (line ~161) uses `innerHTML` to inject HTML content without any sanitization:

```typescript
function textToDOM(text: Text): Node {
  if (typeof text === 'string') {
    return document.createTextNode(text)
  }
  const container = document.createElement('div')
  container.innerHTML = text.html  // unsanitized HTML injection
  const fragment = document.createDocumentFragment()
  while (container.firstChild) {
    fragment.appendChild(container.firstChild)
  }
  return fragment
}
```

The `text.html` value originates from `addOutputSection`, where the `comment` field is constructed via regex replacement on `confidence.comment`. If any part of that data chain carries untrusted or user-influenced content, arbitrary HTML (including `<script>` tags or event handler attributes) could be injected into the DOM.

## Impact

- **Type:** Cross-Site Scripting (XSS)
- **Severity:** Critical
- **Scope:** Playground page — not shipped in the library itself, but the playground is publicly accessible and could be used to demonstrate or exploit XSS if an attacker can control the input.

## Suggested Fix

Either sanitize the HTML before assignment, or build DOM nodes programmatically:

**Option A — Sanitize with an allowlist:**
```typescript
container.innerHTML = DOMPurify.sanitize(text.html)
```

**Option B — Build nodes programmatically (preferred, zero dependencies):**
Replace `innerHTML` usage with safe DOM APIs (`createElement`, `createTextNode`, `appendChild`) to construct the required markup without raw HTML injection.

## References

- [OWASP XSS Prevention Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Scripting_Prevention_Cheat_Sheet.html)

abhishekDeshmukh74:
## Summary

Fixes a cross-site scripting (XSS) vulnerability in the playground's `textToDOM()` function, where unsanitized HTML was injected via `innerHTML`.

Closes #1154

## What changed

**File:** `playground/index.ts`

- **Removed** the `{ html: string }` variant from the `Text` type and the `innerHTML` assignment in `textToDOM()`.
- **Added** `linkifyConfidenceComment()` — a dedicated function that programmatically builds anchor (`<a>`) elements using `createElement` / `createTextNode` instead of parsing raw HTML strings.
- **Updated** the `Text` type to `string | DocumentFragment`, and simplified `textToDOM()` accordingly.
- **Added** `rel="noopener noreferrer"` to the generated external links for defense in depth.

## Why

The previous implementation used `container.innerHTML = text.html` to render the confidence comment link. Although the current data source (`confidence.comment`) is internal to the library, the `innerHTML` pattern is inherently unsafe — any future change that introduces user-influenced data into the `{ html: string }` path would create a direct XSS vector. Building DOM nodes programmatically eliminates this class of vulnerability entirely.

## Testing

- `yarn lint` passes with zero warnings/errors.
- `yarn build` succeeds.
- The confidence comment link renders identically in the playground (link text shows `upgrade to Pro: <base-url>`, href includes full URL with query params, opens in a new tab).

## Backward compatibility

This change is limited to the playground (not shipped in the library distribution). No public API is affected.

### DIFF
--- a/playground/index.ts
+++ b/playground/index.ts
@@ -1,7 +1,7 @@
 import * as FingerprintJS from '../src'
 import { errorToObject } from '../src/utils/misc'
 
-type Text = string | { html: string }
+type Text = string | DocumentFragment
 
 async function getVisitorData() {
   const fp = await FingerprintJS.load({ debug: true })
@@ -26,12 +26,7 @@ async function startPlayground() {
       output,
       header: 'Confidence score:',
       content: String(confidence.score),
-      comment: confidence.comment && {
-        html: confidence.comment.replace(
-          /(upgrade\s+to\s+pro\s+:\s+)?(\s+version)?((https?:\/\/[\w\\.\\/]+)?(\?\S+))?/gi,
-          '<a href="$3" target="_blank">$1$4</a>',
-        ),
-      },
+      comment: confidence.comment && linkifyConfidenceComment(confidence.comment),
       size: 'big',
     })
     addOutputSection({ output, header: 'User agent:', content: navigator.userAgent })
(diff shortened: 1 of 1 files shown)


Valve:

PR #1159, opened 2026-04-12
fix: clamp audio rendering timeout delay to minimum of 0

### LINKED ISSUE #1158: Bug: Potential negative `setTimeout` delay in `startRunningTimeout()` causes premature rejection
### Scenario

In `src/sources/audio.ts`, the `startRunningTimeout` helper computes its delay as:

```typescript
const startRunningTimeout = () => {
  setTimeout(
    () => reject(makeInnerError(InnerErrorName.Timeout)),
    Math.min(runningMaxAwaitTime, startedRunningAt + runningSufficientTime - Date.now()),
  )
}
```

If `Date.now()` is called when the current time has already exceeded `startedRunningAt + runningSufficientTime` (e.g. the JS event loop was busy, the tab was backgrounded briefly, or the device is under heavy load), the computed delay is **negative**.

The [HTML spec](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#dom-settimeout) clamps negative delays to 0 ms, so `setTimeout` fires on the very next macrotask — **immediately** from the caller's perspective. This means the audio rendering promise can be rejected as "timed-out" even when the `OfflineAudioContext` was about to complete successfully.

### Reproduction conditions

- Device or browser under heavy CPU load.
- Tab backgrounded immediately after `load()` is called.
- Any environment where a significant amount of time elapses between the audio context entering the `running` state and `startRunningTimeout()` being called.

### Affected code

**File:** `src/sources/audio.ts` (line ~136)

```typescript
Math.min(runningMaxAwaitTime, startedRunningAt + runningSufficientTime - Date.now())
//       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
//       Can be negative when Date.now() > startedRunningAt + runningSufficientTime
```

### Expected behaviour

The delay should never be negative. A value of `0` is the correct lower bound — it still allows rendering to complete if it is already finished, and avoids the false-positive immediate rejection.

### Suggested fix

Wrap the expression with `Math.max(0, ...)`:

```typescript
const startRunningTimeout = () => {
  setTimeout(
    () => reject(makeInnerError(InnerErrorName.Timeout)),
    Math.max(0, Math.min(runningMaxAwaitTime, startedRunningAt + runningSufficientTime - Date.now())),
  )
}
```

### FingerprintJS version

abhishekDeshmukh74:
### What does this PR do?

Fixes a logic bug in `src/sources/audio.ts` where the `setTimeout` delay inside `startRunningTimeout()` could be **negative**, causing the audio rendering promise to be rejected prematurely.

**Root cause:**

```typescript
// Before — delay can be negative
Math.min(runningMaxAwaitTime, startedRunningAt + runningSufficientTime - Date.now())
```

When `Date.now()` exceeds `startedRunningAt + runningSufficientTime` (e.g. the JS thread was busy, the tab was briefly backgrounded, or the device is under heavy CPU load), the arithmetic yields a negative number. The [HTML spec](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#dom-settimeout) clamps negative delays to 0 ms, so `setTimeout` fires on the very next macrotask — **immediately** — causing a false-positive `Timeout` rejection even when the `OfflineAudioContext` was about to finish successfully.

**Fix:**

Clamp the delay to a minimum of `0` with `Math.max`:

```diff
- Math.min(runningMaxAwaitTime, startedRunningAt + runningSufficientTime - Date.now())
+ Math.max(0, Math.min(runningMaxAwaitTime, startedRunningAt + runningSufficientTime - Date.now()))
```

A delay of `0` is the correct lower bound — it still respects the `runningSufficientTime` budget and avoids the false-positive rejection.

### Changes

- `src/sources/audio.ts` — one-line change wrapping the timeout delay in `Math.max(0, ...)`.

### Related issue

Closes https://github.com/fingerprintjs/fingerprintjs/issues/1158

### Checklist

### DIFF
--- a/src/sources/audio.ts
+++ b/src/sources/audio.ts
@@ -133,7 +133,7 @@ function startRenderingAudio(context: OfflineAudioContext) {
     const startRunningTimeout = () => {
       setTimeout(
         () => reject(makeInnerError(InnerErrorName.Timeout)),
-        Math.min(runningMaxAwaitTime, startedRunningAt + runningSufficientTime - Date.now()),
+        Math.max(0, Math.min(runningMaxAwaitTime, startedRunningAt + runningSufficientTime - Date.now())),
       )
     }
 


Valve:

PR #1161, opened 2026-04-12
fix: replace __proto__ with Object.getPrototypeOf() in getConstantsFromPrototype()

### LINKED ISSUE #1160: Bug: `__proto__` access in `getConstantsFromPrototype()` is non-standard and breaks on null-prototype objects
### Scenario

`getConstantsFromPrototype()` in `src/sources/webgl.ts` accesses `__proto__` directly to enumerate the prototype's keys:

```typescript
function getConstantsFromPrototype<K>(obj: K): Array<Extract<keyof K, string>> {
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  const keys = Object.keys((obj as any).__proto__) as Array<keyof K>
  return keys.filter(isConstantLike)
}
```

This has three problems:

1. **`__proto__` is legacy / non-standard** — it is defined in Annex B of the ECMAScript spec (i.e. a legacy feature for web compatibility) and is not guaranteed to exist in all environments.
2. **Breaks on null-prototype objects** — if `obj` was created with `Object.create(null)`, accessing `.__proto__` returns `undefined`, and `Object.keys(undefined)` throws a `TypeError`.
3. **Hidden by an `as any` cast** — TypeScript cannot catch the hazard because the unsafe access is guarded by `as any`.

The standard, spec-compliant replacement is `Object.getPrototypeOf(obj)`.

### Affected code

**File:** `src/sources/webgl.ts` (line ~229)

```typescript
const keys = Object.keys((obj as any).__proto__) as Array<keyof K>
```

### Expected behaviour

Prototype key enumeration should use `Object.getPrototypeOf()`, guard against a `null` prototype, and not rely on a deprecated legacy accessor.

### Suggested fix

```typescript
function getConstantsFromPrototype<K>(obj: K): Array<Extract<keyof K, string>> {
  const proto = Object.getPrototypeOf(obj) as K | null
  if (!proto) {
    return []
  }
  const keys = Object.keys(proto) as Array<keyof K>
  return keys.filter(isConstantLike)
}
```

### FingerprintJS version

abhishekDeshmukh74:
### What does this PR do?

Replaces the legacy `__proto__` access in `getConstantsFromPrototype()` (`src/sources/webgl.ts`) with the spec-compliant `Object.getPrototypeOf()`.

**Before:**

```typescript
function getConstantsFromPrototype<K>(obj: K): Array<Extract<keyof K, string>> {
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  const keys = Object.keys((obj as any).__proto__) as Array<keyof K>
  return keys.filter(isConstantLike)
}
```

Problems with the old code:
1. **`__proto__` is non-standard** — it is Annex B (legacy web compatibility) and not guaranteed in all environments.
2. **Crashes on null-prototype objects** — `(Object.create(null) as any).__proto__` is `undefined`; `Object.keys(undefined)` throws `TypeError`.
3. **Hidden by `as any`** — TypeScript cannot surface the hazard.

**After:**

```typescript
function getConstantsFromPrototype<K>(obj: K): Array<Extract<keyof K, string>> {
  const proto = Object.getPrototypeOf(obj) as K | null
  if (!proto) {
    return []
  }
  const keys = Object.keys(proto) as Array<keyof K>
  return keys.filter(isConstantLike)
}
```

`Object.getPrototypeOf()` is the standard ES5+ API for prototype introspection. The null-guard ensures the function degrades gracefully for null-prototype objects instead of throwing.

### Changes

- `src/sources/webgl.ts` — replace `__proto__` access with `Object.getPrototypeOf()` and add null-prototype guard.

### Related issue

Closes https://github.com/fingerprintjs/fingerprintjs/issues/1160

### Checklist

### DIFF
--- a/src/sources/webgl.ts
+++ b/src/sources/webgl.ts
@@ -225,8 +225,11 @@ function getShaderPrecision(
 }
 
 function getConstantsFromPrototype<K>(obj: K): Array<Extract<keyof K, string>> {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  const keys = Object.keys((obj as any).__proto__) as Array<keyof K>
+  const proto = Object.getPrototypeOf(obj) as K | null
+  if (!proto) {
+    return []
+  }
+  const keys = Object.keys(proto) as Array<keyof K>
   return keys.filter(isConstantLike)
 }
 


Valve:

PR #1165, opened 2026-04-12
fix: use a monotonic boolean to guard watchScreenFrame() re-entry

### LINKED ISSUE #1164: Bug: `watchScreenFrame()` guard resets to `undefined` after frame is found, allowing redundant re-entry
### Scenario

`watchScreenFrame()` in `src/sources/screen_frame.ts` uses **`screenFrameSizeTimeoutId`** as its "already started" guard:

```typescript
function watchScreenFrame() {
  if (screenFrameSizeTimeoutId !== undefined) return  // guard
  const checkScreenFrame = () => {
    const frameSize = getCurrentScreenFrame()
    if (isFrameSizeNull(frameSize)) {
      screenFrameSizeTimeoutId = (setTimeout as typeof window.setTimeout)(checkScreenFrame, screenFrameCheckInterval)
    } else {
      screenFrameBackup = frameSize
      screenFrameSizeTimeoutId = undefined   // ← guard reset to undefined
    }
  }
  checkScreenFrame()
}
```

The guard has two failure modes:

1. **False re-entry after frame is found.** When `checkScreenFrame` discovers a non-null frame, it sets `screenFrameSizeTimeoutId = undefined` and stores the result. At that point the guard is `undefined` again. Any subsequent call to `watchScreenFrame()` (e.g. from a second `load()` invocation or a future `getUnstableScreenFrame()` call) passes the guard and begins a new, redundant poll loop — even though `screenFrameBackup` is already populated.

2. **Conceptual mismatch.** A timer ID is a transient value that happens to be `undefined` in both the "never started" and the "already finished" states. Using it as a lifecycle guard conflates two very different states, making the intent harder to reason about and leaving the door open for subtle bugs when the wider module is modified.

### Suggested fix

Replace the timer-ID guard with a separate monotonic boolean `watchStarted` that is set to `true` the first time `watchScreenFrame()` runs and **never resets to `false`** during normal operation (only `resetScreenFrameWatch()` touches it, which is test-only):

```typescript
let watchStarted = false

function watchScreenFrame() {
  if (watchStarted) return
  watchStarted = true
  const checkScreenFrame = () => {
    const frameSize = getCurrentScreenFrame()
    if (isFrameSizeNull(frameSize)) {
      screenFrameSizeTimeoutId = (setTimeout as typeof window.setTimeout)(checkScreenFrame, screenFrameCheckInterval)
    } else {
      screenFrameBackup = frameSize
      screenFrameSizeTimeoutId = undefined
    }
  }
  checkScreenFrame()
}

export function resetScreenFrameWatch(): void {
  if (screenFrameSizeTimeoutId !== undefined) {
    clearTimeout(screenFrameSizeTimeoutId)
    screenFrameSizeTimeoutId = undefined
  }
  screenFrameBackup = undefined
  watchStarted = false   // reset for test isolation
}
```

This is a minimal, backward-compatible change: the polling and backup logic are untouched; only the entry guard is made reliably monotonic.

### FingerprintJS version

abhishekDeshmukh74:
### What does this PR do?

Fixes a guard-logic bug in `watchScreenFrame()` (`src/sources/screen_frame.ts`) where the "already started" check could be bypassed after the watch cycle completed, allowing a redundant poll loop to start.

**Root cause:**

The guard relied on `screenFrameSizeTimeoutId !== undefined`:

```typescript
function watchScreenFrame() {
  if (screenFrameSizeTimeoutId !== undefined) return  // guard
  ...
  const checkScreenFrame = () => {
    ...
    } else {
      screenFrameBackup = frameSize
      screenFrameSizeTimeoutId = undefined   // ← guard resets here
    }
  }
}
```

Once a non-null frame is found, `screenFrameSizeTimeoutId` is cleared back to `undefined` and `screenFrameBackup` is populated. Any subsequent call to `watchScreenFrame()` at that point (e.g. from a second `load()` call or a future `getUnstableScreenFrame()` invocation) passes the guard and starts a new, redundant poll loop — even though the work is already done.

**Fix:**

Introduce a separate `screenFrameWatchStarted` boolean that is set to `true` on first entry and **never resets to `false`** during normal operation:

```diff
+ let screenFrameWatchStarted = false

 function watchScreenFrame() {
-  if (screenFrameSizeTimeoutId !== undefined) {
+  if (screenFrameWatchStarted) {
     return
   }
+  screenFrameWatchStarted = true
   ...
 }

 export function resetScreenFrameWatch(): void {
   ...
   screenFrameBackup = undefined
+  screenFrameWatchStarted = false   // reset for test isolation only
 }
```

The polling and backup logic are entirely unchanged. Only the entry guard is made reliably monotonic.

### Changes

- `src/sources/screen_frame.ts` — add `screenFrameWatchStarted` boolean; replace timer-ID guard in `watchScreenFrame()`; reset the boolean in `resetScreenFrameWatch()`.

### Related issue

Closes https://github.com/fingerprintjs/fingerprintjs/issues/1164

### Checklist

### DIFF
--- a/src/sources/screen_frame.ts
+++ b/src/sources/screen_frame.ts
@@ -22,6 +22,7 @@ const roundingPrecision = 10
 // The type is readonly to protect from unwanted mutations
 let screenFrameBackup: Readonly<FrameSize> | undefined
 let screenFrameSizeTimeoutId: number | undefined
+let screenFrameWatchStarted = false
 
 /**
  * Starts watching the screen frame size. When a non-zero size appears, the size is saved and the watch is stopped.
@@ -31,9 +32,10 @@ let screenFrameSizeTimeoutId: number | undefined
  * See more on this at https://github.com/fingerprintjs/fingerprintjs/issues/568
  */
 function watchScreenFrame() {
-  if (screenFrameSizeTimeoutId !== undefined) {
+  if (screenFrameWatchStarted) {
     return
   }
+  screenFrameWatchStarted = true
   const checkScreenFrame = () => {
     const frameSize = getCurrentScreenFrame()
     if (isFrameSizeNull(frameSize)) {
@@ -55,6 +57,7 @@ export function resetScreenFrameWatch(): void {
     screenFrameSizeTimeoutId = undefined
   }
   screenFrameBackup = undefined
+  screenFrameWatchStarted = false
 }
 
 /**


Valve:

PR #1167, opened 2026-04-12
fix: cap total audio render retries and clear pending timer on finalize

### LINKED ISSUE #1166: Bug: unbounded retry loop and leaked timer in `startRenderingAudio()` when tab stays hidden
### Scenario

`startRenderingAudio()` in `src/sources/audio.ts` has two related resource-leak patterns:

#### 1. Unbounded retry loop when the tab is hidden

When `context.state === 'suspended'` and `document.hidden === true`, the foreground retry counter is deliberately **not** incremented (the comment explains this is intentional to wait patiently for the tab to come back into focus). However, the hard-abort condition only checks the foreground count:

```typescript
case 'suspended':
  if (!document.hidden) {
    renderTryCount++   // only incremented in foreground
  }
  if (isFinalized && renderTryCount >= renderTryMaxCount) {
    reject(makeInnerError(InnerErrorName.Suspended))
  } else {
    setTimeout(tryRender, renderRetryDelay)   // ← always schedules another retry
  }
  break
```

If the tab **stays hidden indefinitely**, `renderTryCount` never reaches `renderTryMaxCount`, so `setTimeout(tryRender, renderRetryDelay)` fires forever. Each retry schedules the next one with no upper bound.

#### 2. Pending retry `setTimeout` is never cleared

The return value of `setTimeout(tryRender, renderRetryDelay)` is discarded. If the promise resolves (via `context.oncomplete`) or is rejected by another path while a retry is still pending in the timer queue, the pending callback fires anyway after the fact and calls `context.startRendering()` on an already-completed context.

The `finalize()` function (called on the happy path) has no way to cancel a pending retry timer.

### Impact

- Pathological case: tab hidden indefinitely → retry loop runs until the page is closed, keeping the `OfflineAudioContext`, its promise chain, and all closure variables alive.
- After successful rendering: a queued retry can call `startRendering()` on a completed context, triggering a browser error.

### Suggested fix

1. Add a **hard total-retry cap** (`renderTotalTryMaxCount`) that increments on every retry regardless of `document.hidden`, so the loop terminates even in background.
2. **Store the pending retry timer ID** and clear it inside `finalize()` so a running-path resolution can cancel a background retry that has not yet fired.

```typescript
function startRenderingAudio(context: OfflineAudioContext) {
  const renderTryMaxCount = 3
  const renderTotalTryMaxCount = renderTryMaxCount * 3  // absolute cap including background retries
  const renderRetryDelay = 500
  ...

  const resultPromise = new Promise<AudioBuffer>((resolve, reject) => {
    let isFinalized = false
    let renderTryCount = 0
    let totalRenderTryCount = 0
    let retryTimeoutId: ReturnType<typeof setTimeout> | undefined
    ...

    const tryRender = () => {
      retryTimeoutId = undefined
      try {
        ...
        case 'suspended':
          if (!document.hidden) {
            renderTryCount++
          }
          totalRenderTryCount++
          if (
            (isFinalized && renderTryCount >= renderTryMaxCount) ||
            totalRenderTryCount >= renderTotalTryMaxCount
          ) {
            reject(makeInnerError(InnerErrorName.Suspended))
          } else {
            retryTimeoutId = (setTimeout as typeof window.setTimeout)(tryRender, renderRetryDelay)
          }
          break
      ...

    finalize = () => {
      if (!isFinalized) {
        isFinalized = true
        if (retryTimeoutId !== undefined) {
          clearTimeout(retryTimeoutId)
          retryTimeoutId = undefined
        }
        if (startedRunningAt > 0) {
          startRunningTimeout()
        }
      }
    }
  })
}
```

### FingerprintJS version

abhishekDeshmukh74:
### What does this PR do?

Fixes two related resource-leak issues in `startRenderingAudio()` (`src/sources/audio.ts`).

---

#### Problem 1 — Unbounded retry loop when tab is hidden

When `context.state === 'suspended'` and `document.hidden === true`, the foreground retry counter `renderTryCount` is intentionally not incremented (the comment explains this is by design — the context waits patiently for the tab to come back). However the abort condition only checks the foreground count:

```typescript
if (isFinalized && renderTryCount >= renderTryMaxCount) {
  reject(...)
} else {
  setTimeout(tryRender, renderRetryDelay)   // always retries
}
```

If the tab **stays hidden indefinitely**, `renderTryCount` never reaches `renderTryMaxCount`, so the retry loop runs forever, keeping the `OfflineAudioContext`, its entire promise chain, and all closure variables alive until the page is closed.

**Fix:** Add `renderTotalTryMaxCount = renderTryMaxCount * 3` and a `totalRenderTryCount` counter that always increments regardless of visibility. The abort condition now also checks this absolute ceiling:

```typescript
totalRenderTryCount++
if (
  (isFinalized && renderTryCount >= renderTryMaxCount) ||
  totalRenderTryCount >= renderTotalTryMaxCount
) {
  reject(makeInnerError(InnerErrorName.Suspended))
} else {
  retryTimeoutId = (setTimeout as typeof window.setTimeout)(tryRender, renderRetryDelay)
}
```

---

#### Problem 2 — Pending retry timer is never cleared

The return value of `setTimeout(tryRender, renderRetryDelay)` was discarded. If the promise resolved via `context.oncomplete` (or was rejected by another path) while a retry was still queued, the pending callback would fire anyway and call `context.startRendering()` on an already-completed context.

**Fix:** Store the pending timer ID in `retryTimeoutId`. Clear it at the top of each `tryRender()` invocation (the timer has now fired, so it's no longer pending), and also inside `finalize()` so a successful render cancels any outstanding background retry:

```typescript
finalize = () => {
  if (!isFinalized) {
    isFinalized = true
    if (retryTimeoutId !== undefined) {
      clearTimeout(retryTimeoutId)
      retryTimeoutId = undefined
    }
    if (startedRunningAt > 0) {
      startRunningTimeout()
    }
  }
}
```

---

### Changes

- `src/sources/audio.ts`
  - Add `renderTotalTryMaxCount` constant.
  - Add `totalRenderTryCount` and `retryTimeoutId` variables.
  - Increment `totalRenderTryCount` on every suspended retry.
  - Enforce the absolute cap.
  - Store and clear the retry timer ID.

### Related issue

Closes https://github.com/fingerprintjs/fingerprintjs/issues/1166

### Checklist

- [x] Change is backward compatible (background fingerprinting may now time out with `Suspended` if retried 9+ times; this is strictly better than an unbounded loop)

### DIFF
--- a/src/sources/audio.ts
+++ b/src/sources/audio.ts
@@ -118,6 +118,7 @@ function doesBrowserPerformAntifingerprinting() {
  */
 function startRenderingAudio(context: OfflineAudioContext) {
   const renderTryMaxCount = 3
+  const renderTotalTryMaxCount = renderTryMaxCount * 3 // absolute cap including background retries
   const renderRetryDelay = 500
   const runningMaxAwaitTime = 500
   const runningSufficientTime = 5000
@@ -126,6 +127,8 @@ function startRenderingAudio(context: OfflineAudioContext) {
   const resultPromise = new Promise<AudioBuffer>((resolve, reject) => {
     let isFinalized = false
     let renderTryCount = 0
+    let totalRenderTryCount = 0
+    let retryTimeoutId: ReturnType<typeof setTimeout> | undefined
     let startedRunningAt = 0
 
     context.oncomplete = (event) => resolve(event.renderedBuffer)
@@ -138,6 +141,7 @@ function startRenderingAudio(context: OfflineAudioContext) {
     }
 
     const tryRender = () => {
+      retryTimeoutId = undefined
       try {
         const renderingPromise = context.startRendering()
 
@@ -166,10 +170,14 @@ function startRenderingAudio(context: OfflineAudioContext) {
             if (!document.hidden) {
               renderTryCount++
             }
-            if (isFinalized && renderTryCount >= renderTryMaxCount) {
+            totalRenderTryCount++
+            if (
+              (isFinalized && renderTryCount >= renderTryMaxCount) ||
+              totalRenderTryCount >= renderTotalTryMaxCount
+            ) {
               reject(makeInnerError(InnerErrorName.Suspended))
             } else {
-              setTimeout(tryRender, renderRetryDelay)
+              retryTimeoutId = (setTimeout as typeof window.setTimeout)(tryRender, renderRetryDelay)
             }
             break
         }
@@ -183,6 +191,10 @@ function startRenderingAudio(context: OfflineAudioContext) {
     finalize = () => {
       if (!isFinalized) {
         isFinalized = true
+        if (retryTimeoutId !== undefined) {
+          clearTimeout(retryTimeoutId)
+          retryTimeoutId = undefined
+        }
         if (startedRunningAt > 0) {
           startRunningTimeout()
         }


Valve:

1. It doesn't have automated tests.
2. It doesn't provide a reproducible example (please see my [comment](https://github.com/fingerprintjs/fingerprintjs/issues/1166#issuecomment-4239194003))


1. a reproducible example is provided
2. you add automated tests that demonstrate the fix in a test-driven way.

PR #1169, opened 2026-04-12
fix: replace deprecated execCommand('copy') with Clipboard API in playground

### LINKED ISSUE #1168: Deprecation: replace `document.execCommand('copy')` with Clipboard API in playground
### Scenario

The `copy()` helper in `playground/index.ts` uses the deprecated `document.execCommand('copy')` API:

```typescript
function copy(text: string) {
  const textarea = document.createElement('textarea')
  textarea.value = text
  document.body.appendChild(textarea)
  textarea.focus()
  textarea.select()
  try {
    document.execCommand('copy')
  } catch {
    // Do nothing in case of a copying error
  }
  document.body.removeChild(textarea)
}
```

`document.execCommand` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand) and browser vendors are in the process of removing it. Modern browsers already emit deprecation warnings in DevTools and future browser versions may stop supporting it entirely.

The recommended replacement is the asynchronous [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API) (`navigator.clipboard.writeText()`), which:
- Works in all modern browsers on secure (HTTPS) origins.
- Does not require injecting a hidden `<textarea>` into the DOM.
- Returns a `Promise` so failures can be handled cleanly.

The old `execCommand` path should be kept as a fallback for environments where the Clipboard API is unavailable (e.g. HTTP origins or older browsers).

### Suggested fix

```typescript
async function copy(text: string) {
  if (navigator.clipboard?.writeText) {
    try {
      await navigator.clipboard.writeText(text)
    } catch {
      // Do nothing in case of a copying error
    }
  } else {
    // Legacy fallback for browsers that don't support the Clipboard API
    const textarea = document.createElement('textarea')
    textarea.value = text
    document.body.appendChild(textarea)
    textarea.focus()
    textarea.select()
    try {
      document.execCommand('copy')
    } catch {
      // Do nothing in case of a copying error
    }
    document.body.removeChild(textarea)
  }
}
```

### FingerprintJS version

abhishekDeshmukh74:
### What does this PR do?

Replaces the deprecated `document.execCommand('copy')` call in the `copy()` helper (`playground/index.ts`) with the modern [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API), keeping the old path as a fallback.

**Before:**

```typescript
function copy(text: string) {
  const textarea = document.createElement('textarea')
  textarea.value = text
  document.body.appendChild(textarea)
  textarea.focus()
  textarea.select()
  try {
    document.execCommand('copy')
  } catch {
    // Do nothing in case of a copying error
  }
  document.body.removeChild(textarea)
}
```

`document.execCommand` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand). Browsers already print deprecation warnings and may remove support in future versions.

**After:**

```typescript
async function copy(text: string) {
  if (navigator.clipboard?.writeText) {
    try {
      await navigator.clipboard.writeText(text)
    } catch {
      // Do nothing in case of a copying error
    }
  } else {
    // Legacy fallback for browsers that don't support the Clipboard API
    const textarea = document.createElement('textarea')
    textarea.value = text
    document.body.appendChild(textarea)
    textarea.focus()
    textarea.select()
    try {
      document.execCommand('copy')
    } catch {
      // Do nothing in case of a copying error
    }
    document.body.removeChild(textarea)
  }
}
```

- Modern browsers on HTTPS use `navigator.clipboard.writeText()`.
- The `execCommand` path is preserved in the `else` branch for HTTP origins and older browsers — no regression for any previously-supported environment.
- The function is now `async` to correctly `await` the Clipboard API promise; callers that don't await it continue to work (fire-and-forget is fine for a copy button).

### Changes

- `playground/index.ts` — `copy()` converted to `async`, Clipboard API used as primary path, `execCommand` kept as fallback.

### Related issue

Closes https://github.com/fingerprintjs/fingerprintjs/issues/1168

### Checklist

### DIFF
--- a/playground/index.ts
+++ b/playground/index.ts
@@ -120,18 +120,27 @@ function initializeDebugButtons(debugText: string) {
   }
 }
 
-function copy(text: string) {
-  const textarea = document.createElement('textarea')
-  textarea.value = text
-  document.body.appendChild(textarea)
-  textarea.focus()
-  textarea.select()
-  try {
-    document.execCommand('copy')
-  } catch {
-    // Do nothing in case of a copying error
+async function copy(text: string) {
+  if (navigator.clipboard?.writeText) {
+    try {
+      await navigator.clipboard.writeText(text)
+    } catch {
+      // Do nothing in case of a copying error
+    }
+  } else {
+    // Legacy fallback for browsers that don't support the Clipboard API
+    const textarea = document.createElement('textarea')
+    textarea.value = text
+    document.body.appendChild(textarea)
+    textarea.focus()
+    textarea.select()
+    try {
+      document.execCommand('copy')
+    } catch {
+      // Do nothing in case of a copying error
+    }
+    document.body.removeChild(textarea)
   }
-  document.body.removeChild(textarea)
 }
 
 async function share(text: string) {


playground/index.ts
```diff
@@ -120,18 +120,27 @@ function initializeDebugButtons(debugText: string) {
...
+async function copy(text: string) {
+  if (navigator.clipboard?.writeText) {
+    try {
+      await navigator.clipboard.writeText(text)
+    } catch {
+      // Do nothing in case of a copying error
+    }
+  } else {
+    // Legacy fallback for browsers that don't support the Clipboard API
+    const textarea = document.createElement('textarea')
```
Valve, line 132: this doesn't replace the deprecated execCommand('copy'), it adds a new method that exists alongside the old method.

Valve:
Thanks for the PR!
I think the title of the pull request is misleading, because it doesn't replace the execCommand('copy'), but instead adds a secondary method. 
This adds additional maintenance complexity w/out clear benefits.
I'm closing this PR, but please feel free to re-open it when you think it's possible to **replace** the copy command with the clipboard API. 
If you think that it's possible now, according to the [officially supported browsers](https://docs.fingerprint.com/docs/browser-and-device-support), please feel free to reopen as well, thanks!

PR #1201, opened 2026-08-07
Agent/django fingerprintjs

Midhun884:


primarch42:
Please check our contribution guidelines:
https://github.com/fingerprintjs/fingerprintjs/blob/master/contributing.md

We should not introduce changes of this scope without first opening a discussion or issue and aligning on the approach.

That said, there also seems to be something off with the PR itself, as it contains too many disruptive changes for this type of contribution.

I’m closing the PR for now.

PR #1205, opened 2026-09-10
chore: add stricter lint rules and husky pre-commit hook

primarch42:
Adds stricter ESLint rules and a pre-commit hook to keep code consistent.

#### Lint rules

- Sort imports alphabetically and forbid duplicate imports
- Require `===` / `!==` (no loose equality)
- Prefer template literals over string concatenation
- Disallow unnecessary `async` (functions without `await`)
- Forbid rest destructuring (keeps the bundle smaller)
- Blank line around function definitions
- Warn on focused/skipped tests (`fit`, `fdescribe`, `xit`, `xdescribe`) so they don't get committed by accident

Existing violations were auto-fixed; two were fixed by hand. No behavior changes — `build` and lint pass clean.

#### Pre-commit hook

Husky + lint-staged run `eslint --fix` on staged files before each commit, so lint issues are caught locally instead of in CI.

#### Notes

Default exports are kept — the public API relies on them, so `import/no-default-export` is intentionally not enabled.


resources/content_blocking/make_selectors_tester.ts
```diff
@@ -2,10 +2,13 @@
...
  */
 
-import * as path from 'path'
 import { promises as fsAsync } from 'fs'
+import * as path from 'path'
+
 import * as rollup from 'rollup'
+
 import rollupConfig from '../../rollup.config'
+
```
montenegrino91, line 11: In this file we have
```
import * as path from 'path'

import * as rollup from 'rollup'
```

and in next one we have 
```
import * as fs from 'fs'
import * as readline from 'readline'
```

which is strange to have 2 different behaviours for the same rule, can you please check why is this the case?

montenegrino91 (approved):
All in all it looks good, please just check https://github.com/fingerprintjs/fingerprintjs/pull/1205/changes#r3977139186 prior to merging, thanks.

resources/content_blocking/make_selectors_tester.ts
igor-piskunov-fp: If I got this right, the difference is between:

```
builtin
<black line>
external
```

and 

```
builtin
builtin
```

igor-piskunov-fp: Yeah, here it is

Image

montenegrino91: I do get that part, but it is bit odd to have it separated and not grouped although it has the same "pattern". But since it is only bugging me, we can dismiss my comment 🤷🏽

